PR #1, opened 2017-11-20
Integrate PHPStan analysis

ondrejmirtes:


build.xml
```diff
@@ -95,4 +96,19 @@
...
+
+	<target name="phpstan">
+		<exec
+			executable="vendor/bin/phpstan"
+			logoutput="true"
+			passthru="true"
+			checkreturn="true"
+		>
+			<arg value="analyse"/>
+			<arg value="-l"/>
```
JanTvrdik: Options shortcuts are for humans. When written to bash scripts / build tools, the full option name should always be used for readability, i.e. `--level` in this case.

ondrejmirtes: @JanTvrdik Updated.

PR #3, opened 2017-12-06
added new method "getNodes" in PhpDocNode.

AmsTaFFix:
this method returns all nodes from phpdoc node. It can be helpfull to get access to all nodes, like PhpDocTextNode


JanTvrdik:
The `$children` property is public, so there is not much need for a getter

AmsTaFFix:
oh my god! I'm so sorry, I so unaccustomed to public fields... I see only private and protected :) I'm sorry, and thank you for answer

PR #8, opened 2018-03-27
[.git] Add dev files to export-ignore

Slamdunk:


TomasVotruba:
What is your motivation for this PR?

Slamdunk:
> What is your motivation for this PR?

Targeted IDE scan of project source

TomasVotruba:
What does that mean exactly?

Slamdunk:
When I search for the usages of some source, i.e. a class, I get all the sources from the `tests/` folder, which are useless if I'm **using** the library. Those hints are useful only if I'm **developing** this library, and if so I'm cloning it instead of downloading its zip file.

I'm supposing that we all know that composer installs libraries as zip by default, which are built on Github skipping files and directories marked as `export-ignore` in this PR file.

TomasVotruba:
I see. I never came to such problem myself, so I was curious. 
I probably get my source classes with higher priority in the list.

JanTvrdik:
@Slamdunk Thanks!

PR #23, opened 2019-03-18
Support deprecated tags

mglaman:
It seems like this would make handling multiline deprecated messages easier, like Throws or Return tags https://github.com/phpstan/phpstan/pull/1983

### DIFF
--- a/src/Parser/PhpDocParser.php
+++ b/src/Parser/PhpDocParser.php
@@ -99,6 +99,10 @@ class PhpDocParser
 					$tagValue = $this->parseThrowsTagValue($tokens);
 					break;
 
+				case '@deprecated':
+					$tagValue = $this->parseDeprecatedTagValue($tokens);
+					break;
+
 				case '@property':
 				case '@property-read':
 				case '@property-write':
@@ -157,6 +161,12 @@ class PhpDocParser
 		$type = $this->typeParser->parse($tokens);
 		$description = $this->parseOptionalDescription($tokens, true);
 		return new Ast\PhpDoc\ThrowsTagValueNode($type, $description);
+	}
+
+	private function parseDeprecatedTagValue(TokenIterator $tokens): Ast\PhpDoc\DeprecatedTagValueNode
+	{
+		$description = $this->parseOptionalDescription($tokens);
+		return new Ast\PhpDoc\DeprecatedTagValueNode($description);
 	}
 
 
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Hi, this looks okay but a test would be nice :) We can then update 1983.

mglaman:
Yes, I will do that. Had to run before I could add, but wanted to get something up before I forgot 😄

mglaman:
I have added some tests. I wonder if this will help handle deprecations in PHPStan and the messaging. Hopefully :)

JanTvrdik:
I don't understand what is this PR for. Parsing `@deprecated` tags is already supported. I does not need any special code.

mglaman:
Technically they can be discovered. Other tags have typed objects. I was hoping this would improve parsing of deprecated messages in PHPStan itself per my PR linked previously

ondrejmirtes:
I think that if there are other types of nodes in the switch then there should be `@deprecated` as well - simplifies the usage for phpdoc-parser users.

ondrejmirtes:
Please fix the build. Now all jobs fail. (Failure on 7.3 is fine, I need to update PHPCS on master.)

JanTvrdik:
> Other tags have typed objects

It already has a typed object (`GenericTagValueNode`).

> I was hoping this would improve parsing of deprecated messages 

It uses the same parsing the default branch.

src/Parser/PhpDocParser.php
```diff
@@ -159,6 +163,12 @@ private function parseThrowsTagValue(TokenIterator $tokens): Ast\PhpDoc\ThrowsTa
 		return new Ast\PhpDoc\ThrowsTagValueNode($type, $description);
 	}
 
+	private function parseDeprecatedTagValue(TokenIterator $tokens): Ast\PhpDoc\DeprecatedTagValueNode
+	{
+		$description = $this->parseOptionalDescription($tokens, true);
```
JanTvrdik: the second parameter should be `false`

ondrejmirtes:
`DeprecatedTagValueNode` is a little bit more type-safe but I understand it's a small gain. Anyway, I wouldn't want to block custom parsing for this because other planned tags like `@template` will need it too.

mglaman:
@ondrejmirtes I was finally able to circle back to this now that DrupalCon is over. I have PHP 7.3 on all of my locals, so took a little bit to fix PHPCS but it's all looking green again.

mglaman:
@ondrejmirtes 🙌 thanks!

ondrejmirtes:
I rebased, made some changes and merged it :) For example the description shouldn't have been nullable, it was never null besides the test case.

mglaman:
Thanks! Appreciate that, helping me learn a few things.

PR #26, opened 2019-04-24
Enhance support for multiple line descriptions

mglaman:
I didn't want to touch \PHPStan\PhpDocParser\Parser\TokenIterator, which I think could be refactored to properly implement the \Iterator interface. That seemed like a task of its own.

This loops over the tokens passed to `\PHPStan\PhpDocParser\Parser\PhpDocParser::parseDeprecatedTagValue` and builds the entire message.

### DIFF
--- a/src/Parser/PhpDocParser.php
+++ b/src/Parser/PhpDocParser.php
@@ -90,7 +90,7 @@ class PhpDocParser
 			}
 
 			// There's more text on a new line, ensure spacing.
-			$text .= ' ';
+			$text .= "\n";
 		}
 		$text = trim($text, " \t");
 
--- a/src/Parser/PhpDocParser.php
+++ b/src/Parser/PhpDocParser.php
@@ -170,7 +170,18 @@ class PhpDocParser
 
 	private function parseDeprecatedTagValue(TokenIterator $tokens): Ast\PhpDoc\DeprecatedTagValueNode
 	{
-		$description = $this->parseOptionalDescription($tokens);
+		$description = '';
+		while ($tokens->currentTokenType() !== Lexer::TOKEN_CLOSE_PHPDOC) {
+			$description .= $tokens->joinUntil(Lexer::TOKEN_PHPDOC_EOL, Lexer::TOKEN_CLOSE_PHPDOC, Lexer::TOKEN_END);
+			$description = rtrim($description, " \t");
+			if ($tokens->currentTokenType() !== Lexer::TOKEN_PHPDOC_EOL) {
+				break;
+			}
+			// There's more text on a new line, ensure spacing.
+			$description .= ' ';
+			$tokens->next();
+		}
+		$description = rtrim($description, " \t");
 		return new Ast\PhpDoc\DeprecatedTagValueNode($description);
 	}
 
--- a/src/Parser/PhpDocParser.php
+++ b/src/Parser/PhpDocParser.php
@@ -81,9 +81,6 @@ class PhpDocParser
 			// to be combined.
 			$tokens->pushSavePoint();
 			$tokens->next();
-			if ($tokens->currentTokenType() === Lexer::TOKEN_PHPDOC_EOL) {
-				$tokens->next();
-			}
 			if ($tokens->currentTokenType() !== Lexer::TOKEN_IDENTIFIER) {
 				$tokens->rollback();
 				break;
--- a/src/Parser/PhpDocParser.php
+++ b/src/Parser/PhpDocParser.php
@@ -61,8 +61,38 @@ class PhpDocParser
 
 	private function parseText(TokenIterator $tokens): Ast\PhpDoc\PhpDocTextNode
 	{
-		$text = $tokens->joinUntil(Lexer::TOKEN_PHPDOC_EOL, Lexer::TOKEN_CLOSE_PHPDOC, Lexer::TOKEN_END);
-		$text = rtrim($text, " \t"); // the trimmed characters MUST match Lexer::TOKEN_HORIZONTAL_WS
+		$text = '';
+		while (true) {
+			// If we received a Lexer::TOKEN_PHPDOC_EOL, exit early to prevent
+			// them from being processed.
+			if ($tokens->currentTokenType() === Lexer::TOKEN_PHPDOC_EOL) {
+				break;
+			}
+			$text .= $tokens->joinUntil(Lexer::TOKEN_PHPDOC_EOL, Lexer::TOKEN_CLOSE_PHPDOC, Lexer::TOKEN_END);
+			$text = rtrim($text, " \t");
+
+			// If we joined until TOKEN_PHPDOC_EOL, peak at the next tokens to see
+			// if we have a multiline string to join.
+			if ($tokens->currentTokenType() !== Lexer::TOKEN_PHPDOC_EOL) {
+				break;
+			}
+
+			// Peek at the next token to determine if it is more text that needs
+			// to be combined.
+			$tokens->pushSavePoint();
+			$tokens->next();
+			if ($tokens->currentTokenType() === Lexer::TOKEN_PHPDOC_EOL) {
+				$tokens->next();
+			}
+			if ($tokens->currentTokenType() !== Lexer::TOKEN_IDENTIFIER) {
+				$tokens->rollback();
+				break;
+			}
+
+			// There's more text on a new line, ensure spacing.
+			$text .= ' ';
+		}
+		$text = trim($text, " \t");
 
 		return new Ast\PhpDoc\PhpDocTextNode($text);
 	}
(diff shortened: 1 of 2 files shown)


mglaman:
I've added a test case which matches a failure from https://github.com/phpstan/phpstan/pull/1983 as well.

JanTvrdik:
This needs to be implemented for all tags or for none (i.e. modify `parseOptionalDescription` method).
The parsing will fail, if the `@deprecated` is NOT the last tag, e.g. 

```php
		yield [
			'OK with long descriptions',
			'/**
			  * @deprecated foo
			  * @deprecated bar
			  */',
			new PhpDocNode([
				new PhpDocTagNode(
					'@deprecated',
					new DeprecatedTagValueNode('foo')
				),
				new PhpDocTagNode(
					'@deprecated',
					new DeprecatedTagValueNode('bar')
				),
			]),
		];
```



---

> which I think could be refactored to properly implement the \Iterator interface

Yes, but it would be mostly useless. `TokenIterator` methods provides much finer and precise control than `Iterator` methods. Or is there a benefit I'm missing?

ondrejmirtes:
I'll wait for you guys to resolve this, @JanTvrdik is the expert in this area :)

mglaman:
@JanTvrdik in regards to iterator: it just felt hard to have a controlled loop over the tokens. I was constantly pushing the internal index out of bounds and having errors thrown. Iterator would allow us to have valid.

It already implements most of the interface's methods, that is why I suggested

mglaman:
> The parsing will fail, if the @deprecated is NOT the last tag, e.g.

I see the failure

```
'description' => 'text first @deprecated text second'
```

So the fix should go in parseOptionalDescription itself and not generic for deprecated.

JanTvrdik:
> it just felt hard to have a controlled loop over the tokens

But how would `Iterator` interface helps? You would still be unable to use `foreach` (because it triggers `rewind`).

> I see the failure

I don't understand what you mean. Tags which are in the middle of line are intentionally ignored.

mglaman:
> But how would Iterator interface help? You would still be unable to use foreach (because it triggers rewind).

Yeah, I see it doesn't really help. Was looking to leverage a `valid` method, but didn't need to.

@JanTvrdik that error was a test result and the current implementation was consuming the next line, forcing the second deprecation to be marked as part of the first.

I've made a push that puts the logic in `parseText` to handle multiple lined texts.

tests/PHPStan/Parser/PhpDocParserTest.php
```diff
@@ -1518,10 +1576,9 @@ public function provideMultiLinePhpDocData(): array
 							new IdentifierTypeNode('Foo'),
 							false,
 							'$foo',
-							'1st multi world description'
+							'1st multi world description some text in the middle'
```
mglaman: This has no additional line breaks, so it is considered a continuation of the parameter description.

tests/PHPStan/Parser/PhpDocParserTest.php
```diff
@@ -1538,15 +1595,16 @@ public function provideMultiLinePhpDocData(): array
 				'/**
 				  *
 				  *
-				  * @param Foo $foo 1st multi world description
+				  * @param Foo $foo 1st multi world description with empty lines
```
mglaman: Appended to help differentiate to catch an in-development bug where it was consuming all empty line breaks between the next `Lexer::TOKEN_IDENTIFIER`

src/Parser/PhpDocParser.php
```diff
@@ -60,8 +60,38 @@ private function parseChild(TokenIterator $tokens): Ast\PhpDoc\PhpDocChildNode
...
+			if ($tokens->currentTokenType() === Lexer::TOKEN_PHPDOC_EOL) {
+				$tokens->next();
+			}
+			if ($tokens->currentTokenType() !== Lexer::TOKEN_IDENTIFIER) {
+				$tokens->rollback();
+				break;
+			}
+
+			// There's more text on a new line, ensure spacing.
+			$text .= ' ';
```
JanTvrdik: The lines must be separated by `\n`.

tests/PHPStan/Parser/PhpDocParserTest.php
JanTvrdik: the empty lines must be retained (because they are significant when phpdoc contains e.g. markdown)

JanTvrdik:
Few more things needs to be done.

1. compare the parsing logic with phpDocumentor
2. compare the parsing logic with PhpStorm
3. test it against large real-world [dataset](https://github.com/phpstan/phpdoc-parser/files/3120312/phpdoc.zip) (extracted phpdoc from top ~1000 packages on packagist) with focus on how real people write multi-line descriptions (how do they format it, where do they put new lines and indentations...)

mglaman:
@JanTvrdik I have no problem doing that. I made a commit which takes a lengthy example from a Drupal class, and we heavily rely on phpDoc for our docs. https://github.com/phpstan/phpdoc-parser/pull/26/commits/aba3906c61801f3be3eef0b6586c365a5a7f0858. Documentation: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Field%21FieldItemInterface.php/function/FieldItemInterface%3A%3Aschema/8.6.x

To be honest, I feel like that is now bleeding outside the scope of this initial PR as no existing test coverage ever handled `@return` or `@param` with descriptions on a new line (the drupalcs standard.)

mglaman:
For what its worth: I'm fine working with what needs to be done. But per my last commit, it seems a generic baseline test needs to be made against `master` for comparison.

goba:
The title of this pull request should be updated to the latest/current scope. I tried to verify how phpDocumentor parses multiline phpdoc, although I don't expect any surprises there. After all this is just about concatenating all the items... However I only found compiler pass implementations (at https://github.com/phpDocumentor/phpDocumentor2/tree/develop/src/phpDocumentor/Compiler/Pass) and could not find where the actual parsing happens.

JanTvrdik:
@goba I think it's better to just that it than trying to read the code.

src/Parser/PhpDocParser.php
mglaman: But the intention is for a sentence to be combined if they were broken onto new lines due to width. It assumes a single line break is a continuation and double line breaks are a new entry. This is how phpDocs are parsed.

mglaman:
@JanTvrdik I have added an example and would like your feedback. It highlights other problems in the phpdoc parser itself compared to expectations. I need to know if the scope of this PR is changing or not, because I need to decided if I should change route on using PHPStan or not.

src/Parser/PhpDocParser.php
JanTvrdik: > It assumes a single line break is a continuation and double line breaks are a new entry.

Sure, but do people writing phpDocs assume the same? This for example prevents the description from being processed as markdown (because in markdown newline is significant). I think the newlines should be kept as-is. A higher-level library may always choose to replace single newlines with space.

goba: https://stackoverflow.com/questions/11741302/how-do-i-put-blocks-of-php-code-into-a-phpdoc-docblock/40036952#40036952 cites various phpdoc variant sources and all suggest to support Markdown indeed in single line and multiline tags. Eg. a code example would use markdown and have significant newlines.

Drupal uses the apparantely non-standard @code tag where the content should also be significant.

Here is a Drupal example from core/core.api.php:

```
 * This will be an object of class \Drupal\Core\Config\Config, which has methods
 * for getting configuration information. For instance, if your YAML file
 * structure looks like this:
 * @code
 * enabled: '0'
 * bar:
 *   baz: 'string1'
 *   boo: 34
 * @endcode
```

Markdown examples on the link :)

mglaman: I understand the situation better, now. I'll add the `\n`.

mglaman: @JanTvrdik this was addressed

mglaman:
I did some experimenting with https://github.com/phpDocumentor/ReflectionDocBlock, and it does preserve `\n` per @JanTvrdik recommendation before.  In https://github.com/phpstan/phpstan/pull/1983 we can choose to convert into a single line if we choose (as you've suggested.)

@325fc6414ee15a704a9644211d7c9b1cd560145d fixes problems with lost new lines, bringing it back to match previous parsing and phpDoc.

mglaman:
I have always retitled the PR to be more accurate :)

mglaman:
@ondrejmirtes @JanTvrdik  this has been rebased against latest master

ondrejmirtes:
It's perfect to my eyes, thank you! I'll merge it, test it and tag it right away!

mglaman:
❤️ thanks!

tests/PHPStan/Parser/PhpDocParserTest.php
```diff
@@ -2302,4 +2368,179 @@ public function provideTemplateTagsData(): \Iterator
 		];
 	}
 
+	public function providerDebug(): \Iterator
```
staabm, line 2371: while [adding `shipmonk/dead-code-detector` to the codebase](https://github.com/phpstan/phpdoc-parser/pull/315) it detected that this provider is never registered with any test.

I wonder how this method is supposed to work, or whether its an oversight that this method is not referenced anywhere?

(sorry for commenting on this multi-year old PR, but I figured it would be best to get back to the people involved in the change itself)

PR #30, opened 2019-06-03
Array shapes support

arnaud-lb:
Adds support for parsing array shapes (object-like arrays).

Examples:

```
array{'foo': int, "bar": string}
array{0: int, 1?: int}
array{int, int}
array{foo: int, bar: string}
```

Keys can be single quoted strings, double quoted strings, integers, or identifiers (represented as ConstExprStringNode, ConstExprIntegerNode, IdentifierTypeNode).

They can be marked optional with a `?` before the key and the `:`.

They can be omitted entirely.

Values can be any type, including nullable types or nested array shapes.

### DIFF
--- a/doc/grammars/type.abnf
+++ b/doc/grammars/type.abnf
@@ -52,6 +52,12 @@ CallableReturnType
 Array
 	= 1*(TokenSquareBracketOpen TokenSquareBracketClose)
 
+ArrayShape
+    = TokenCurlyBracketOpen ArrayShapeItem *(TokenComma ArrayShapeItem) TokenCurlyBracketClose
+
+ArrayShapeItem
+    = (ConstantString / ConstantInt / TokenIdentifier) TokenNullable TokenColon Type
+    / Type
 
 ; ---------------------------------------------------------------------------- ;
 ;  ConstantExpr                                                                ;
@@ -138,6 +144,12 @@ TokenSquareBracketOpen
 
 TokenSquareBracketClose
 	= "]" *ByteHorizontalWs
+
+TokenCurlyBracketOpen
+    = "{" *ByteHorizontalWs
+
+TokenCurlyBracketClose
+    = "}" *ByteHorizontalWs
 
 TokenComma
 	= "," *ByteHorizontalWs
--- a/src/Ast/Type/ArrayShapeNode.php
+++ b/src/Ast/Type/ArrayShapeNode.php
@@ -1,0 +1,22 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast\Type;
+
+class ArrayShapeNode implements TypeNode
+{
+
+	/** @var ArrayShapeItemNode[] */
+	public $items;
+
+	public function __construct(array $items)
+	{
+		$this->items = $items;
+	}
+
+
+	public function __toString(): string
+	{
+		return 'array{' . implode(', ', $this->items) . '}';
+	}
+
+}
--- a/src/Parser/TypeParser.php
+++ b/src/Parser/TypeParser.php
@@ -253,19 +253,11 @@ class TypeParser
 	}
 
 	/**
-	 * @return Ast\ConstExpr\ConstExprStringNode|Ast\ConstExpr\ConstExprIntegerNode|Ast\Type\IdentifierTypeNode
+	 * @return Ast\ConstExpr\ConstExprIntegerNode|Ast\Type\IdentifierTypeNode
 	 */
 	private function parseArrayShapeKey(TokenIterator $tokens)
 	{
-		if ($tokens->isCurrentTokenType(Lexer::TOKEN_SINGLE_QUOTED_STRING)) {
-			$key = new Ast\ConstExpr\ConstExprStringNode($tokens->currentTokenValue());
-			$tokens->next();
-
-		} elseif ($tokens->isCurrentTokenType(Lexer::TOKEN_DOUBLE_QUOTED_STRING)) {
-			$key = new Ast\ConstExpr\ConstExprStringNode($tokens->currentTokenValue());
-			$tokens->next();
-
-		} elseif ($tokens->isCurrentTokenType(Lexer::TOKEN_INTEGER)) {
+		if ($tokens->isCurrentTokenType(Lexer::TOKEN_INTEGER)) {
 			$key = new Ast\ConstExpr\ConstExprIntegerNode($tokens->currentTokenValue());
 			$tokens->next();
 
--- a/src/Parser/TypeParser.php
+++ b/src/Parser/TypeParser.php
@@ -54,7 +54,7 @@ class TypeParser
 			} elseif ($tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_SQUARE_BRACKET)) {
 				$type = $this->tryParseArray($tokens, $type);
 
-			} elseif ($type->name === 'array' && $tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_CURLY_BRACKET)) {
+			} elseif ($type->name === 'array' && $tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_CURLY_BRACKET) && !$tokens->isPrecededByHorizontalWhitespace()) {
 				$type = $this->parseArrayShape($tokens, $type);
 			}
 		}
@@ -97,7 +97,7 @@ class TypeParser
 		if ($tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_ANGLE_BRACKET)) {
 			$type = $this->parseGeneric($tokens, $type);
 
-		} elseif ($type->name === 'array' && $tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_CURLY_BRACKET)) {
+		} elseif ($type->name === 'array' && $tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_CURLY_BRACKET) && !$tokens->isPrecededByHorizontalWhitespace()) {
 			$type = $this->parseArrayShape($tokens, $type);
 		}
 
@@ -174,7 +174,7 @@ class TypeParser
 			if ($tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_ANGLE_BRACKET)) {
 				$type = $this->parseGeneric($tokens, $type);
 
-			} elseif ($type->name === 'array' && $tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_CURLY_BRACKET)) {
+			} elseif ($type->name === 'array' && $tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_CURLY_BRACKET) && !$tokens->isPrecededByHorizontalWhitespace()) {
 				$type = $this->parseArrayShape($tokens, $type);
 			}
 		}
--- a/src/Parser/TypeParser.php
+++ b/src/Parser/TypeParser.php
@@ -53,6 +53,9 @@ class TypeParser
 
 			} elseif ($tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_SQUARE_BRACKET)) {
 				$type = $this->tryParseArray($tokens, $type);
+
+			} elseif ($type->name === 'array' && $tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_CURLY_BRACKET)) {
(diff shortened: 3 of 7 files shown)


arnaud-lb:
Related: https://github.com/phpstan/phpstan/issues/2173

ondrejmirtes:
I love it! 😍 But looks like you've forgot to add those new classes in. The build (after my CS fix) fails on:

```
1) Warning
The data provider specified for PHPStan\PhpDocParser\Parser\TypeParserTest::testParse is invalid.
Class 'PHPStan\PhpDocParser\Ast\Type\ArrayShapeNode' not found
```

BTW: The optional keys in PHPStan are right now modeled with a union of two constant arrays - with the keys and without the keys. Unfortunately, this leads to some wrongly interpreted code. We can improve that later for example by introducing optional keys to ConstantArrayType but that brings more work. I'd continue represent them with unions of two arrays right now.

src/Parser/TypeParser.php
```diff
@@ -53,6 +55,9 @@ private function parseAtomic(TokenIterator $tokens): Ast\Type\TypeNode
 
 			} elseif ($tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_SQUARE_BRACKET)) {
 				$type = $this->tryParseArray($tokens, $type);
+
+			} elseif ($type->name === 'array' && $tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_CURLY_BRACKET)) {
+				$type = $this->parseArrayShape($tokens, $type);
```
JanTvrdik: this should be maybe `tryParseArrayShape`, so that `/** @return array {curly description} */` is not suddenly invalid

src/Parser/TypeParser.php
```diff
@@ -208,4 +219,63 @@ private function tryParseArray(TokenIterator $tokens, Ast\Type\TypeNode $type):
 		return $type;
 	}
 
+
+	private function parseArrayShape(TokenIterator $tokens, Ast\Type\TypeNode $type): Ast\Type\TypeNode
+	{
+		$tokens->consumeTokenType(Lexer::TOKEN_OPEN_CURLY_BRACKET);
+		$items = [$this->parseArrayShapeItem($tokens)];
+
+		while (!$tokens->tryConsumeTokenType(Lexer::TOKEN_CLOSE_CURLY_BRACKET)) {
```
JanTvrdik: for consistency with `parseGeneric` a (micro) speed improvements, this should be written as

```php
while ($tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA)) {
  $items[] = $this->parseArrayShapeItem($tokens);
}

$tokens->consumeTokenType(Lexer::TOKEN_CLOSE_CURLY_BRACKET);
```

src/Parser/TypeParser.php
```diff
@@ -208,4 +219,63 @@ private function tryParseArray(TokenIterator $tokens, Ast\Type\TypeNode $type):
...
+			$value = $this->parse($tokens);
+
+			return new Ast\Type\ArrayItemNode(null, $optional, $value);
+		}
+	}
+
+	/**
+	 * @return ConstExprStringNode|ConstExprIntegerNode|IdentifierTypeNode
+	 */
+	private function parseArrayShapeKey(TokenIterator $tokens)
```
JanTvrdik: missing native type `Ast\Type\Node` (or whatever is the proper name)

arnaud-lb: Are you refering to the return type ? What would you suggest ?

src/Parser/TypeParser.php
arnaud-lb: The problem with `tryParse` is that it silently ignores errors. An array shape specification will be silently ignored if it has a minor typo in it.

Alternatively, if we don't allow spaces between `array` and `{`, this would prevent the problem you mention. We could do that by adding an `array{` token. What do you think ?

src/Parser/TypeParser.php
arnaud-lb: Ok, got it :)

JanTvrdik: yes, imho the only common is `\PHPStan\PhpDocParser\Ast\Node`

src/Parser/TypeParser.php
JanTvrdik: > The problem with tryParse is that it silently ignores errors. 

That's not a problem of `tryParse` but inherent problem of phpDoc which is designed to be human friendly and not computer friendly (same problem as markdown).

JanTvrdik: without `tryParse` it's a BC break.

src/Parser/TypeParser.php
```diff
@@ -208,4 +219,64 @@ private function tryParseArray(TokenIterator $tokens, Ast\Type\TypeNode $type):
...
+		}
+	}
+
+	/**
+	 * @return ConstExprStringNode|ConstExprIntegerNode|Ast\Type\IdentifierTypeNode
+	 */
+	private function parseArrayShapeKey(TokenIterator $tokens)
+	{
+		if ($tokens->isCurrentTokenType(Lexer::TOKEN_SINGLE_QUOTED_STRING)) {
+			$key = new ConstExprStringNode($tokens->currentTokenValue());
```
JanTvrdik: for consistency should be `Ast\ConstExpr\ConstExprStringNode`

src/Parser/TypeParser.php
```diff
@@ -208,4 +219,64 @@ private function tryParseArray(TokenIterator $tokens, Ast\Type\TypeNode $type):
...
+			$tokens->next();
+
+		} elseif ($tokens->isCurrentTokenType(Lexer::TOKEN_DOUBLE_QUOTED_STRING)) {
+			$key = new ConstExprStringNode($tokens->currentTokenValue());
+			$tokens->next();
+
+		} elseif ($tokens->isCurrentTokenType(Lexer::TOKEN_INTEGER)) {
+			$key = new ConstExprIntegerNode($tokens->currentTokenValue());
```
JanTvrdik: for consistency should be `Ast\ConstExpr\ConstExprIntegerNode`

src/Ast/Type/ArrayItemNode.php
```diff
@@ -0,0 +1,45 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast\Type;
+
+use PHPStan\PhpDocParser\Ast\ConstExpr\ConstExprIntegerNode;
+use PHPStan\PhpDocParser\Ast\ConstExpr\ConstExprStringNode;
+
+class ArrayItemNode implements TypeNode
```
JanTvrdik: should be imho `ArrayShapeTypeItemNode` or maybe just `ShapeTypeItemNode`

src/Ast/Type/ArrayItemNode.php
```diff
@@ -0,0 +1,45 @@
...
+{
+
+	/** @var ConstExprStringNode|ConstExprIntegerNode|IdentifierTypeNode|null */
+	public $keyName;
+
+	/** @var bool */
+	public $keyOptional;
```
JanTvrdik: imho just `$optional` is enough

src/Ast/Type/ArrayShapeNode.php
```diff
@@ -0,0 +1,24 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast\Type;
+
+class ArrayShapeNode implements TypeNode
```
JanTvrdik: should be imho `ArrayShapeTypeNode` or maybe just `ShapeTypeNode`

src/Ast/Type/ArrayShapeNode.php
```diff
@@ -0,0 +1,24 @@
...
+	{
+		$this->items = $items;
+	}
+
+
+	public function __toString(): string
+	{
+		return 'array{' . implode(', ', array_map(static function (ArrayItemNode $node): string {
+			return (string) $node;
+		}, $this->items)) . '}';
```
JanTvrdik: can be simplified as 

```php
return 'array{' . implode(', ', $this->items) . '};
```

doc/grammars/type.abnf
```diff
@@ -52,6 +52,12 @@ CallableReturnType
 Array
 	= 1*(TokenSquareBracketOpen TokenSquareBracketClose)
 
+ArrayShape
+    = TokenCurlyBracketOpen ArrayItem *(TokenComma ArrayItem) TokenCurlyBracketClose
+
+ArrayItem
```
JanTvrdik: `ArrayShapeItem`

JanTvrdik:
I wonder if we should make `array{` byte sequence a new token instead of having `array` and `{` as two tokens. Mostly depends on whether allowing `array   {string, int}` is desired or not.

Also it should be noted that allowing quoted string (which psalm does not allow) introduces grammar inconsistency (because it creates a place which looks like it expects constants expression but identifiers are not treated as constants) so I'm not sure it's a good idea.

JanTvrdik:
cc @muglug

src/Parser/TypeParser.php
arnaud-lb: So what do you think of not allowing a whitespace between `array` and `{` ? This would prevent the BC break, while also avoiding silenced errors (which are not user friendly).

JanTvrdik: Yes, that should be fine.

arnaud-lb:
> I wonder if we should make `array{` byte sequence a new token instead of having `array` and `{` as two tokens. Mostly depends on whether allowing `array {string, int}` is desired or not.

I'm in favor of not allowing whitespaces between `array` and `{` (by introducing a new `array{` token, or by taking profit of the fact that the tokeniser reports whitespace tokens).

Pros:
 - We can report syntax errors in a `array{` definition without BC break
 - No silent errors
Cons:
 - We can not write `array {` (with a whitespace)

> Also it should be noted that allowing quoted string (which psalm does not allow) introduces grammar inconsistency (because it creates a place which looks like it expects constants expression but identifiers are not treated as constants) so I'm not sure it's a good idea.

While looking at Psalm I though it was supported, but you are right: it's not. In Pslam, `array{'a': int}` will expect an array with the key `'\'a\''`, and not `'a'`. I think that there is a risk of generating confusing errors here: `expected array{'a': int}, array{a: int} provided`.

So I would be in favor of either allowing quoted strings (so that `array{'a': int}` expects a key `'a'`), or to explicitly disallow them (such that `array{'a': int}` is a syntax error).

src/Parser/TypeParser.php
arnaud-lb: https://github.com/phpstan/phpdoc-parser/pull/30/commits/03768b2e21f5c0953e394709fc0a018b259fba8b

ondrejmirtes:
Thank you both for improving this!

@arnaud-lb I don't see a testcase for something wrong inside the `array{...}` so that we know you're not using `tryParse...`.

arnaud-lb:
Added :)

More about keys:

- If we allow only identifiers, we limit the set of keys we can express with this syntax to the keys that are valid identifiers. Psalm would support our syntax entirely.
- If we allow identifiers and quoted strings at the same time, there are grammar inconsistencies, as noted by @JanTvrdik.
- If we allow only quoted strings, the grammar inconsistencies disappears, and we can express all possible keys, however we lose any compatibility with Psalm on array shapes.
- Psalm allows anything as key, and treats the whole thing as a string: if there are quotes, they are part of the key's value. All characters appear to be allowed. I think that this is dangerous, as this prevents any backwards-compatible evolution of this syntax in the future. This is also confusing when keys appears to be quoted strings, but are not really.

Thoughts ?

JanTvrdik:
> If we allow only identifiers, we limit the set of keys we can express with this syntax to the keys that are valid identifiers.

I think this is the best option for now.

muglug:
Array key escaping is an open issue https://github.com/vimeo/psalm/issues/1518

Psalm should support it, I just haven’t got around to it yet

ondrejmirtes:
We should have a test that verifies that quoted string as a key parsing fails :) Otherwise it's ready to merge, right, everyone?

arnaud-lb:
👍 added a test

ondrejmirtes:
Thank you very much! I'll start poking it with ConstantArrayType right away :)

ondrejmirtes:
And done: https://github.com/phpstan/phpstan/commit/c4ed3c7c8ded963aa8bec0c6c789a5090998f5c9  Feel free to review :)

PR #31, opened 2019-07-16
Support @extends, @implements, @uses

arnaud-lb:
This adds support for the `@extends`, `@implements`, `@uses` tags. These tags can be used on a class-like to give type information about existing inheritance, implementations, or trait uses.

- `@extends`: Denotes that the class extends the given class with the given parameterized types
- `@implements`: Denotes that the class implements the given interface with the given parameterized types
- `@uses`: Denotes that the class uses the given trait with the given parameterized types

Grammar:

```
'@' ('extends' | 'implements' | 'uses') identifier '<' type (',' type)* '>'
```

Example:

``` php
<?php

/**
 * @template K
 * @template V
 *
 * @implements Collection<K,V>
 */
class Foo implements Collection {
}
```

### DIFF
--- a/src/Ast/PhpDoc/ExtendsTagValueNode.php
+++ b/src/Ast/PhpDoc/ExtendsTagValueNode.php
@@ -1,0 +1,28 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast\PhpDoc;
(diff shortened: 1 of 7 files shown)


ondrejmirtes:
Thank you! 🙌

PR #33, opened 2019-10-22
Support multiple lines for array shape

shmax:
Possible fix for https://github.com/phpstan/phpstan/issues/2497

Hey there, great library, fantastic work to all involved. So, like a lot of folks I'm excited about using array shapes, but without multiline syntax it doesn't really seem practical. 

I haven't done much with parsers since my school days of long ago, and I was told that you have previously dismissed this idea as not feasible, so please tell me what I'm doing wrong.

### DIFF
--- a/src/Ast/PhpDoc/TemplateTagValueNode.php
+++ b/src/Ast/PhpDoc/TemplateTagValueNode.php
@@ -26,7 +26,7 @@ class TemplateTagValueNode implements PhpDocTagValueNode
 
 	public function __toString(): string
 	{
-		$bound = $this->bound ? " of {$this->bound}" : '';
+		$bound = isset($this->bound) ? " of {$this->bound}" : '';
 		return trim("{$this->name}{$bound} {$this->description}");
 	}
 
--- a/src/Ast/PhpDoc/TemplateTagValueNode.php
+++ b/src/Ast/PhpDoc/TemplateTagValueNode.php
@@ -26,7 +26,7 @@ class TemplateTagValueNode implements PhpDocTagValueNode
 
 	public function __toString(): string
 	{
-		$bound = isset($this->bound) ? " of {$this->bound}" : '';
+		$bound = $this->bound !== null ? " of {$this->bound}" : '';
 		return trim("{$this->name}{$bound} {$this->description}");
 	}
 
--- a/src/Parser/TypeParser.php
+++ b/src/Parser/TypeParser.php
@@ -227,7 +227,7 @@ class TypeParser
 		$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
 		while ($tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA)) {
 			$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
-			if($tokens->tryConsumeTokenType(Lexer::TOKEN_CLOSE_CURLY_BRACKET)) {
+			if ($tokens->tryConsumeTokenType(Lexer::TOKEN_CLOSE_CURLY_BRACKET)) {
 				// trailing comma case
 				return new Ast\Type\ArrayShapeNode($items);
 			}
--- a/src/Parser/TypeParser.php
+++ b/src/Parser/TypeParser.php
@@ -221,12 +221,15 @@ class TypeParser
 	private function parseArrayShape(TokenIterator $tokens, Ast\Type\TypeNode $type): Ast\Type\TypeNode
 	{
 		$tokens->consumeTokenType(Lexer::TOKEN_OPEN_CURLY_BRACKET);
+		$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
 		$items = [$this->parseArrayShapeItem($tokens)];
 
 		while ($tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA)) {
+			$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
 			$items[] = $this->parseArrayShapeItem($tokens);
 		}
 
+		$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
 		$tokens->consumeTokenType(Lexer::TOKEN_CLOSE_CURLY_BRACKET);
 
 		return new Ast\Type\ArrayShapeNode($items);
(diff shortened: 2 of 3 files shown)


ondrejmirtes:
Only @JanTvrdik or @arnaud-lb are able to review this 😊

src/Ast/PhpDoc/TemplateTagValueNode.php
```diff
@@ -26,7 +26,7 @@ public function __construct(string $name, ?TypeNode $bound, string $description)
 
 	public function __toString(): string
 	{
-		$bound = $this->bound ? " of {$this->bound}" : '';
+		$bound = isset($this->bound) ? " of {$this->bound}" : '';
```
JanTvrdik: should be `$this->bound !== null`

tests/PHPStan/Parser/TypeParserTest.php
```diff
@@ -393,7 +393,12 @@ public function provideParseData(): array
 				]),
 			],
 			[
-				'array{a: int, b: array{c: callable(): int}}',
+				'array{
+				    a: int, 
+				    b: array{
+				        c: callable(): int
+				    }
+				 }',
```
JanTvrdik: You should add new test instead of modifying an existing one

src/Ast/PhpDoc/TemplateTagValueNode.php
```diff
@@ -26,7 +26,7 @@ public function __construct(string $name, ?TypeNode $bound, string $description)
 
 	public function __toString(): string
 	{
-		$bound = $this->bound ? " of {$this->bound}" : '';
```
arnaud-lb: We should check for `null` explicitly :)

src/Parser/TypeParser.php
```diff
@@ -221,12 +221,15 @@ private function tryParseArray(TokenIterator $tokens, Ast\Type\TypeNode $type):
...
+		$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
 		$items = [$this->parseArrayShapeItem($tokens)];
 
 		while ($tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA)) {
+			$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
 			$items[] = $this->parseArrayShapeItem($tokens);
 		}
 
+		$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
 		$tokens->consumeTokenType(Lexer::TOKEN_CLOSE_CURLY_BRACKET);
```
arnaud-lb: I think that we should also allow line breaks before the comma, so that this works:

```
array{
    int
    , int
    , int
}
```

We might want to allow trailing commas as well:


```
array{
    int,
    int,
    int,
}
```

tests/PHPStan/Parser/TypeParserTest.php
JanTvrdik: Also the test is incorrect, as it's missing the `*`. What you are actually parsing is sth like

```
array{
 * a: int,
 * b: array{
 *   c: callable(): int
 * }
 * }
```

arnaud-lb: During real usage, the lexer will receive this with a `* ` prefix on each line. We should add the prefix on this input, so that the test can detect any regressions related to this in the future.

src/Parser/TypeParser.php
ondrejmirtes: I agree. Trailing commas are awesome :)

ondrejmirtes: Each of these examples should have a separate test case.

ondrejmirtes:
If we're able to finish this and merge it, I'm gonna release next phpdoc-parser in the current 0.3.x series and PHPStan 0.11.19 with support for multiline array shapes! 👍

src/Ast/PhpDoc/TemplateTagValueNode.php
shmax: Fixed in 7e0ad3a

tests/PHPStan/Parser/TypeParserTest.php
shmax: Is it all right if I just add a single test showing that the leading `*`s don't disrupt anything? I'd otherwise like to omit them for readability. 414daf3

src/Parser/TypeParser.php
shmax: Line breaks before commas 1f26308

shmax:
Did my best to add the requested features. I feel like the `parseArrayShape` method could be cleaned up a bit; something like

```
while (!$tokens->tryConsumeTokenType(Lexer::TOKEN_CLOSE_CURLY_BRACKET)) {
			$tokens->consumeTokenType(Lexer::TOKEN_COMMA);
			$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
			$items[] = $this->parseArrayShapeItem($tokens);
			$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
		}
```

But then it just sort of runs wild and doesn't throw when an expected closing curly isn't encountered. Any suggestions?

Also, what's the best way to run lint and stan on this project? I didn't see a `scripts` section in composer.json, so I've just been running `phpcs` directly...

ondrejmirtes:
You can always check what `.travis.yml` does :) It's `vendor/bin/phing`.

shmax:
> You can always check what .travis.yml does :) It's vendor/bin/phing.

That's a bit more comprehensive, thank you. The `/tools/abnfgen` file doesn't build in my environment, but I assume that's because I'm using windows.

ondrejmirtes:
I'm happy with this. If someone has any more comments, we can fix them on master :) Thank you very much!

shmax:
Yee-haw, thanks much. Looking forward to the next release...

shmax:
@ondrejmirtes By the way, can you shed any light on what is going on with slevomat/coding-standard? Array-shape support was added in [June](https://github.com/slevomat/coding-standard/commit/35446d2331feab750bb5019212f12057785eacbe), but there hasn't been a release since [March](https://github.com/slevomat/coding-standard/releases/tag/5.0.4) ...

ondrejmirtes:
Nothing bad is going on with Slevomat CS, it’s an actively maintained project. @kukulich is planning to do a release.

shmax:
Very well, then, thank you.

PR #34, opened 2019-11-06
Add @template-covariant support

arnaud-lb:


JanTvrdik:
That seems way too long. Why not extend the existing `@template` annotation to support in/out keywords (Kotlin) or +/- signs (Scala)?

```php
/**
 * @template in T
 */
```

ondrejmirtes:
We need to support what's already in the wild. But we also plan to support `@phpstan-generic<K, V>` syntax and prefer it in the docs etc. See the discussion there (and feel free to contribute to it!): https://github.com/phpstan/phpstan/issues/2527

muglug:
> That seems way too long. Why not extend the existing @template annotation to support in/out keywords (Kotlin) or +/- signs (Scala)?

The reason I went with `-covariant` is that it seemed to me, as a newcomer to the idea, something that should be annotated fairly explicitly. 

Also I felt `out` was confusing given by-ref parameters: https://github.com/vimeo/psalm/issues/1603#issuecomment-489745389

PR #37, opened 2020-01-23
Allow [] after generic type or array shape - closes #36

### LINKED ISSUE #36: Generic<Type>[] incorrectly failing
# Bug report

Something like
```
Collection<int, int>[]
```
Should be a valid type

No error with psalm: https://psalm.dev/r/26d249d13d
An error with phpstan (the last one): https://phpstan.org/r/a42e2d1b-d1fe-4c8b-a4df-9c6e61b422b1

```
PHPDoc tag @param has invalid value (Collection<int, int>[] $bars): Unexpected token "[", expected variable at offset 38
```

An actual workaround is `(Collection<int, int>)[]` but I think the phpDoc parser should still be improved to handle the syntax.

iluuu1994:


iluuu1994:
@JanTvrdik Can you review this too since you raised some concerns?

JanTvrdik:
I hate it. I want the ambiguous `[]` syntax to die.

The most practical mess will be probably with callable types, because `callable(): string[]` will now be parsed as array of callables returning string.

iluuu1994:
Lol, I'll add a test for that to be sure. I still don't see how `[]` are ambiguous. They're not more ambiguous than any other postfix operator. I don't like the fact that it defaults to `TKey` being `array-key` but I don't think the syntax will go anywhere soon.

ondrejmirtes:
> I don't like the fact that it defaults to TKey being array-key but I don't think the syntax will go anywhere soon.

I don't know what you're talking about, can you post an example?

The callable example is bad and should be solved in an elegant way (so that it means "callable that returns array<string>"). But it's still a competetive disadvantage that PHPStan doesn't understand this syntax now, so I'd like to solve it, albeit with compromises.

ondrejmirtes:
I've given it a little bit more thought and the problem with this implementation is that it does more than what the title says:

> Allow [] after generic type or array shape

It tries to parse `[]` after everything, which has the problem of ambiguity. So if we limit it only to generics and array shapes for now, it should be fine?

JanTvrdik:
> I still don't see how [] are ambiguous

It can mean either arra

ondrejmirtes:
Yeah, but we will have to live with that. It's like wishing IPv4 to die...

iluuu1994:
> I don't know what you're talking about, can you post an example?

What I meant is just that I'd prefer `[]` to be a packed array by default, so what Psalm considers to be a [list](https://psalm.dev/docs/annotating_code/type_syntax/array_types/). But that would make 90% of the array annotations in the wild incorrect.

iluuu1994:
> It tries to parse [] after everything

Anything that is atomic. The problem is that callables aren't atomic. There's still a PR from @JanTvrdik to solve that.

ondrejmirtes:
I'll merge this if you do it only for generics and array shapes + write a test case that shows that `callable(): string[]` didn't change :) It has the highest potential of not breaking anything. Thanks.

iluuu1994:
Done. I also noticed the following:

```php
callable(): int[]
// was parsed as callable: int

$this[]
// was parsed as $this

?int[]
// was parsed as ?int, is now array<int>|null analogous to Psalm
```

ondrejmirtes:
Thank you very much!

iluuu1994:
@ondrejmirtes You're welcome 🙂

Issue #42, opened 2020-05-04
Multi Line Generic Types are invalid


ServerExe:
# Bug report

PhpStan-Version: 0.12.19

The following error occurs when trying to declare a multi lined Generics annotation:
```
PHPDoc tag @implements has invalid value (HelloInterface< Foo, Bar): Unexpected token "\n *", expected type at offset 34
```

### Code snippet that reproduces the problem
https://phpstan.org/r/80a149d9-5820-4041-bf3b-4b6e19de6b09

### Expected output
`No errors!` as it is in:
https://phpstan.org/r/08857dad-5e7b-4615-9c99-f3397aa204ce

icanhazstring:
@ondrejmirtes thanks for moving. Can you point ppl in the right direction, so maybe someone can get startet right away :)

ondrejmirtes:
See how multiline array shapes are implemented: https://github.com/phpstan/phpdoc-parser/blob/d4c4cd3caecfa40661cff48b24b6b5216c488458/src/Parser/TypeParser.php#L282-L304

This is where generic types are parsed: https://github.com/phpstan/phpdoc-parser/blob/d4c4cd3caecfa40661cff48b24b6b5216c488458/src/Parser/TypeParser.php#L165-L176

icanhazstring:
Will try to tackle this one :)

icanhazstring:
@ondrejmirtes would it be wise to update the `Tokenzier` to avoid tokenizing the `\n` as a token itself.
The only thing "breaking" the code right now is the "new_line" token. With this removed, everything should work fine.

icanhazstring:
Nevermind, just found out how the parsing works exactly :)

icanhazstring:
This can be closed then @ondrejmirtes :)

ondrejmirtes:
Yes, implemented in https://github.com/phpstan/phpdoc-parser/pull/44 :)

PR #44, opened 2020-05-04
Add multiline generic support

icanhazstring:
Solve #42

### DIFF
--- a/tests/PHPStan/Parser/PhpDocParserTest.php
+++ b/tests/PHPStan/Parser/PhpDocParserTest.php
@@ -3078,6 +3078,54 @@ chunk. Must be higher that in the previous request.'),
 				),
 			]),
 		];
+
+//		yield [
+//			'multiline generic types - trailing comma',
+//			'/**' . PHP_EOL .
+//			' * @implements Foo<' . PHP_EOL .
+//			' *    A,' . PHP_EOL .
+//			' * >' . PHP_EOL .
+//			' */',
+//			new PhpDocNode([
+//				new PhpDocTagNode(
+//					'@implements',
+//					new ImplementsTagValueNode(
+//						new GenericTypeNode(
+//							new IdentifierTypeNode('Foo'),
+//							[
+//								new IdentifierTypeNode('A'),
+//							]
+//						),
+//						''
+//					)
+//				),
+//			]),
+//		];
+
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Parser/PhpDocParserTest.php
```diff
@@ -3046,6 +3046,30 @@ public function provideRealWorldExampleData(): \Iterator
 				),
 			]),
 		];
+
+		yield [
```
ondrejmirtes, line 3050: Please add these tests that check for:

1) That trailing comma after last element is allowed
2) That leading comma at the beginning of the line is also allowed (I know it's ugly but I've seen it in the wild).

Thank you, otherwise 👍

icanhazstring: Ah, sure thing. Will do 👍

icanhazstring: Just a quick question, trailing comma should not work at all am I right?
`array<A, B,>` is not a correct notation

Where a the leading one should be fine, as long as there are two types given
```
array<
   A
   , B
>
```

JanTvrdik: (Kotlin allows trailing comma in generic declarations since 1.4).

ondrejmirtes: Trailing commas are getting popular and even supported by PHP itself. Since this works: https://phpstan.org/r/e1c07184-7f6d-48ba-99cd-29dd41fe5349, I'm also fine if it will work in generics.

ondrejmirtes: We shouldn't support multiple trailing commas: `A,B,,` - that should be an error, otherwise they're fine.

icanhazstring: Thanks for informations. Will add support for them then :)

ondrejmirtes:
Awesome, love it :)

ondrejmirtes:
Released as 0.4.7, but will probably take a few days to be baked into the next release of PHPStan :)

icanhazstring:
Thanks 🥳

PR #50, opened 2020-07-23
⬆️  Upgrade PHP dependency to support PHP8 installation

Th3Mouk:
The release of PHP8 alpha approaching, it will be nice if the lib authorize installation without tweaking composer installation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ php:
   - 7.1
   - 7.2
   - 7.3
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Hi, I prefer ^7.1 || ^8.0 instead.

Th3Mouk:
@ondrejmirtes no problem, fixed

ondrejmirtes:
Thank you!

PR #52, opened 2020-08-30
TypeParser: variadic arguments are always considered optional

dktapps:
this fixes phpstan/phpstan#3798.
it's legal to pass zero arguments to a variadic parameter.
I've ensured that it's still possible to write things like `callable(int ...$foo=) : void` to preserve backwards compatibility, but really it should not be necessary or desirable to add the optional flag to a variadic parameter anyway.


dktapps:
Test failures seem to be caused by some additional faulty assumptions about variadics. I'll attempt to locate them.

ondrejmirtes:
This is wrong place to fix this. The AST should differentiate between `callable(int ...$i): void` and `callable(int ...$i=): void`, no matter the semantics. This should be fixed in TypeNodeResolver in phpstan-src instead. There are already several conditions like this in phpstan-src already.

dktapps:
Got it, I'll dig into this later.

Issue #11, opened 2018-05-05
Allow attributes to make format preserving print piece of cake


TomasVotruba:
It probably out of scope of PHPStan (for now), but as we talked, it would be great if this package allowed *format preserving print*.

I already asked [nikic](https://github.com/nikic/PHP-Parser/issues/487) how to do it, he explained me the basics. Basically, `AbstractNode` class with set/get attributes is the first step towards it: https://github.com/nikic/PHP-Parser/commit/4d2a4d02b08b75d00067f8bf8b3c58a993abc0d0#diff-ee2208021c6e96ff1de44281aa029630R108

After few weeks I managed to make [working prototype](https://github.com/Symplify/Symplify/tree/master/packages/BetterPhpDocParser#better-phpdoc-parser), but it contains lot of boiler plate code and external storage of tokens positions etc., to make it work.

Having this would also make it easy to configure FQN namespaces, the same way [php-parser does](https://github.com/nikic/PHP-Parser/blob/3193f7aae7b6aaa1432081bf3832b01b5da0373d/lib/PhpParser/NodeVisitor/NameResolver.php#L185), without modifing the output. And other SOLID features :)

ondrejmirtes:
Please join forces with @dantleech in #10 - he wrote in two hours ago 😊

TomasVotruba:
I've read it, but Tokens (array, chars) and Nodes (right objects) are different thing

dantleech:
Quite funny, I was just looking at your packages @TomasVotruba  :) My approach for format preserving was to leave the AST read-only, but instead export it to a mutable format, e.g. XML. In the XML document the nodes are represented as elements, and the tokens have node values, e.g.

```
<PhpDocNode>
  <PhpDocTextNode text="Short description"/>
  <PhpDocTextNode text=""/>
  <PhpDocTextNode text="This is a multi-line description"/>
  <PhpDocTextNode text="of the docblock."/>
  <PhpDocTextNode text=""/>
  <PhpDocTagNode name="@param">
    <ParamTagNode isVariadic="" parameterName="$bar" description="">
      <IdentifierTypeNode name="Fobar"/>
    </ParamTagNode>
  </PhpDocTagNode>
  <PhpDocTagNode name="@param">
    <ParamTagNode isVariadic="" parameterName="$foo" description="">
      <IdentifierTypeNode name="Bar\Foo"/>
    </ParamTagNode>
  </PhpDocTagNode>
[...]

PR #57, opened 2021-02-07
Add attributes to each node

TomasVotruba:
Closes https://github.com/phpstan/phpdoc-parser/issues/11

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*.php]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = tab
+indent_size = 4
--- a/src/Ast/BaseNode.php
+++ b/src/Ast/BaseNode.php
@@ -23,16 +23,15 @@ abstract class BaseNode implements Node
 	}
 
 	/**
-	 * @param mixed|null $default
 	 * @return mixed|null
 	 */
-	public function getAttribute(string $key, $default = null)
+	public function getAttribute(string $key)
 	{
 		if ($this->hasAttribute($key)) {
 			return $this->attributes[$key];
 		}
 
-		return $default;
+		return null;
 	}
 
 }
--- a/src/Ast/BaseNode.php
+++ b/src/Ast/BaseNode.php
@@ -1,0 +1,37 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast;
+
+abstract class BaseNode implements Node
+{
+
+	/** @var array<string, mixed> */
+	private $attributes = [];
+
+    /**
+     * @param mixed $value
+     */
+	public function setAttribute(string $key, $value): void
+	{
+		$this->attributes[$key] = $value;
+	}
+
+	public function hasAttribute(string $key): bool
+	{
+		return array_key_exists($key, $this->attributes);
+	}
+
+	/**
+	 * @param mixed|null $default
+	 * @return mixed|null
+	 */
+	public function getAttribute(string $key, $default = null)
+	{
+		if ($this->hasAttribute($key)) {
+			return $this->attributes[$key];
+		}
+
+		return $default;
+	}
+
+}
--- a/src/Ast/BaseNode.php
+++ b/src/Ast/BaseNode.php
@@ -8,9 +8,10 @@ abstract class BaseNode implements Node
 	/** @var array<string, mixed> */
 	private $attributes = [];
 
-    /**
-     * @param mixed $value
-     */
+	/**
+	 * @param string $key
+	 * @param mixed $value
+	 */
 	public function setAttribute(string $key, $value): void
 	{
 		$this->attributes[$key] = $value;
--- a/build-cs/composer.json
+++ b/build-cs/composer.json
@@ -3,8 +3,9 @@
 		"php": "^7.1 || ^8.0"
 	},
 	"require-dev": {
-		"consistence/coding-standard": "^3.10",
+		"consistence-community/coding-standard": "^3.10",
 		"dealerdirect/phpcodesniffer-composer-installer": "^0.7.1",
+		"squizlabs/php_codesniffer": "^3.5",
 		"slevomat/coding-standard": "^6.4.1"
 	}
 }
--- a/build.xml
+++ b/build.xml
@@ -60,7 +60,6 @@
 			<arg value="install"/>
 			<arg value="--working-dir"/>
 			<arg path="build-cs"/>
-			<arg value="--ignore-platform-reqs"/>
 			<arg value="--ansi"/>
 		</exec>
 		<exec
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <ruleset name="PHPStan PHPDoc Parser">
 	<config name="php_version" value="70100"/>
-	<rule ref="build-cs/vendor/consistence/coding-standard/Consistence/ruleset.xml">
+	<rule ref="build-cs/vendor/consistence-community/coding-standard/Consistence/ruleset.xml">
 		<exclude name="SlevomatCodingStandard.Commenting.InlineDocCommentDeclaration.InvalidFormat"/>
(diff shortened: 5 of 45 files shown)


src/Ast/BaseNode.php
```diff
@@ -0,0 +1,34 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast;
+
+abstract class BaseNode implements Node
+{
+
+	/** @var array<string, mixed> */
+	protected $attributes = [];
```
ondrejmirtes, line 9: Should be private

ondrejmirtes (changes requested):
I also expected the parser to fill in those attributes needed for a format-preserving printer :) Right now the nodes will come with zero attributes from the parser.

TomasVotruba:
I'm trying to figure out how CI works here. It seems hidden in some layer that does not allow PHP 8 to run cs :/

.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*.php]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = tab
+indent_size = 4
```
TomasVotruba, line 9: This avoid tabs/spaces conflicts in files out of the box

image

ondrejmirtes: CS is in `/build-cs` and runs only on PHP 7.4 afaik. PHP_CodeSniffer is probably not yet compatible with PHP 8.

src/Ast/BaseNode.php
```diff
@@ -0,0 +1,37 @@
...
+abstract class BaseNode implements Node
+{
+
+	/** @var array<string, mixed> */
+	private $attributes = [];
+
+	/**
+	 * @param mixed $value
+	 */
+	public function setAttribute(string $key, $value): void
```
TomasVotruba, line 14: I don't get this one. What's wrong here?

```bash
FILE: ...ome/runner/work/phpdoc-parser/phpdoc-parser/src/Ast/BaseNode.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 14 | ERROR | Unknown type hint "string" found for $value
    |       | (Squiz.Commenting.FunctionComment.InvalidTypeHint)
----------------------------------------------------------------------
```

.editorconfig
TomasVotruba: Yes :/ 
https://github.com/phpstan/phpdoc-parser/pull/57#issuecomment-774729574

src/Ast/BaseNode.php
```diff
@@ -0,0 +1,37 @@
...
+	public function hasAttribute(string $key): bool
+	{
+		return array_key_exists($key, $this->attributes);
+	}
+
+	/**
+	 * @param mixed|null $default
+	 * @return mixed|null
+	 */
+	public function getAttribute(string $key, $default = null)
```
JanTvrdik, line 28: why do you need the `$default`?

TomasVotruba: Not really, just convention from php-parser. It can be removed

.editorconfig
TomasVotruba: This package is blocker for PHP 8: https://github.com/consistence/coding-standard/blob/4b4b199b14071c26d6b8064a19a4227c1700a000/composer.json#L14

src/Ast/BaseNode.php
```diff
@@ -0,0 +1,37 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast;
+
+abstract class BaseNode implements Node
+{
+
+	/** @var array<string, mixed> */
+	private $attributes = [];
```
JanTvrdik, line 9: Why not just `public $attributes = []` without the methods? All other properties on nodes are public as well.

.editorconfig
TomasVotruba: It seems @vasekPurchart is not maintaining that packages for a while https://github.com/consistence/coding-standard/pull/65

src/Ast/BaseNode.php
ondrejmirtes: I like the convenience methods `get`/`set`/`has`, with public property you'd have juggle with ugly `isset` or `array_key_exists` in userland code...

TomasVotruba: See https://github.com/phpstan/phpdoc-parser/pull/57/files/1647b310b0d93d2b070875d96558e721335caa07#diff-1c468e54ea94bf2ee8fdc56af68ed6c184a4a54c3298ab74baad9b6404c98bf1

Also better type declaration support on method types.
Making it `public` would override everything with assigns, which is not desired on attributes (compared to all other properties on nodes)

src/Ast/BaseNode.php
ondrejmirtes: Probably a bug in an obsolete PHPCS version. I'm gonna look into this once I'm happy with the code and prepare to merge it 👍

.editorconfig
TomasVotruba: Resolved in standalone PR https://github.com/phpstan/phpdoc-parser/pull/58

src/Ast/BaseNode.php
TomasVotruba: See https://github.com/phpstan/phpdoc-parser/pull/58 first

ondrejmirtes: #58 merged, please rebase this PR so we can see it passes :)

src/Ast/BaseNode.php
ondrejmirtes: Yeah, please always return null if the attribute doesn't exist, I don't think we'd want to override it.

ondrejmirtes (changes requested):
Please modify the parser to actually fill in those attributes needed for a format-preserving printer, otherwise it's fine.

TomasVotruba:
Hi Ondra,

life came into way and I have other priorities ATM, so I won't have time to play around printer. I don't have it working in my packages yet either.
But it would be very helpful to allow attributes in the nodes. Rector could drop ~40 classes that it uses to generate attribute-aware nodes and whole code would get clean-up. It would also ease testing for printer.

Would you be ok to merge this PR with attributes only and then later handle the printer? 
If so, I will update the PR with your comments and make CI pass.

ondrejmirtes:
Can you please verify this PR is backwards compatible with your other code? Probably is because you'd just be overriding an existing method, but I wouldn't want to break your existing code by releasing this in 0.4.13.

TomasVotruba:
I've tested it sample and it should be ok: 
https://3v4l.org/D4PkT

If it will break, I'll checkout latest 0.9.x Rector release and relase a new version with compatible code.
I'll be here today so I can patch it promptly.

TomasVotruba:
Rebased on `master` and merged fixup commits

src/Ast/BaseNode.php
TomasVotruba: I will :+1:

TomasVotruba:
Ready to review :heavy_check_mark:

TomasVotruba:
I've just ported this branch to Rector and it works well :+1: Ready to merge for me. 

See https://github.com/rectorphp/rector/pull/5841 (PHPStan is not relevant; it's failing, because now the classes are loaded twice to give this PR a priority),

ondrejmirtes:
Yeah, give me a bit of time, I'm doing big refactoring this weekend (https://twitter.com/OndrejMirtes/status/1370778018566832131) related to handling `static` type, I'll look into this once I'm finished.

TomasVotruba:
Sure, good luck with those last test :+1:  They're the worst :D

ondrejmirtes:
Hi, I decided on a bit different approach: https://github.com/phpstan/phpdoc-parser/pull/65

1) Add the methods to the Node interface
2) Use a trait instead of class inheritance.

I'm still crediting you in the commit. Because of the first change, I'll have to release a new major version (0.5). I believe you'll be able to use the PR very easily anyway (https://github.com/rectorphp/rector/pull/5841). Thanks.

TomasVotruba:
Hey, that's good news :+1: 

I'll try to update my PR in Rector

Issue #59, opened 2021-02-12
Annotation @TYPO3\CMS\Extbase\Annotation\Inject not tokenized correctly?


sabbelasichon:
Currently i have the problem that all annotations prefixed with `@TYPO3` are not tokenized correctly by the Lexer class. The end result of the tag variable is always `@TYPO.` The rest is cut off.
I guess the failure is located here https://github.com/phpstan/phpdoc-parser/blob/master/src/Lexer/Lexer.php#L144 because numbers are not allowed. 
I am not really sure how i can test it or prove it. If you could guide me i would be happy to fix this, if it is considered a bug

JanTvrdik:
I guess we could change the pattern to `@[a-z-][a-z0-9-]*+`

sabbelasichon:
@JanTvrdik Should i open a PR for that?

PR #60, opened 2021-02-15
[TASK] Allow numbers in tags

sabbelasichon:
Relates: #59


ondrejmirtes:
What I need to see is a test for this change, I don't believe one-line regex changes like that :)

sabbelasichon:
@ondrejmirtes: Total agree. At least the existing one did not break. But what could be a real test case? I thought we are going to discuss this. I came up only with a contrived examples. Because the "official" ones don´t have numbers in it.

ondrejmirtes:
Add another test in this file https://github.com/phpstan/phpdoc-parser/blob/master/tests/PHPStan/Parser/PhpDocParserTest.php, it has to fail without this regex modification.

PR #62, opened 2021-02-27
Support for type alias tags

jiripudil:
This PR adds support for [`@psalm-type`](https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-type) and [`@psalm-import-type`](https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-import-type) as well as `@phpstan-` prefixed alternatives:

```php
/**
 * @phpstan-type TypeAlias callable(string): bool
 */
class Foo {}

/**
 * @phpstan-import-type TypeAlias from Foo as ImportedAlias
 */
class Bar {}
```

### DIFF
--- a/src/Ast/PhpDoc/PhpDocNode.php
+++ b/src/Ast/PhpDoc/PhpDocNode.php
@@ -238,6 +238,20 @@ class PhpDocNode implements Node
 	}
 
 
+	/**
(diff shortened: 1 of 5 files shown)


ondrejmirtes:
This is perfect! Thank you.

PR #66, opened 2021-03-19
Fixed "Undefined array key -1"

kukulich:


ondrejmirtes:
Thank you! Released as 0.4.14 and 0.5.1.

PR #68, opened 2021-03-20
Fix PhpDocNode end newline print

TomasVotruba:
This has to be worked around on every `PhpDocNode` in Rector: https://github.com/rectorphp/rector/blob/57acbe7cdca1bb73e155920c5f9596bd5f10e2b6/packages/AttributeAwarePhpDoc/Ast/PhpDoc/AttributeAwarePhpDocNode.php#L22

This should fix it here:

## Before 

```php
/**
 * @var int*/
```

## After

```php
/**
 * @var int
 */
```

### DIFF
--- a/src/Ast/PhpDoc/PhpDocNode.php
+++ b/src/Ast/PhpDoc/PhpDocNode.php
@@ -271,7 +271,7 @@ class PhpDocNode implements Node
 
 	public function __toString(): string
 	{
-		return "/**\n * " . implode("\n * ", $this->children) . '*/';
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Please add a test.

TomasVotruba:
The `PhpDocParserTest` only compared the 2 nodes are printed same way, which is always true.
I've added a new test file, so the printed node is compared to expected string.

ondrejmirtes:
Thank you!

PR #70, opened 2021-04-03
type aliases: more defensive parsing of type alias import targets

jiripudil:
This makes sure that `$importedFrom` is always actually `IdentifierTypeNode`, and catches the most obvious unsupported types such as

```php
/**
 * @phpstan-import-type Foo from 42
 * @phpstan-import-type Foo from ArraysAreNotSupported[]
 * @phpstan-import-type Foo from NeitherAreGenerics<string>
 */
```

It has been suggested (https://github.com/phpstan/phpstan/issues/3001#issuecomment-754272815) that local aliases could somehow support generics. As far as I am concerned, that's still far in the future and will need some careful consideration, so I'm proposing to restrict it now as well, and perhaps revisit it later once it's more clear how generic aliases should work.

### DIFF
--- a/src/Parser/PhpDocParser.php
+++ b/src/Parser/PhpDocParser.php
@@ -403,8 +403,8 @@ class PhpDocParser
 			);
 		}
 
-		$importedFrom = $this->typeParser->parse($tokens);
-		assert($importedFrom instanceof IdentifierTypeNode);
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you!

Issue #71, opened 2021-04-08
Update .gitattributes


williamdes:
https://github.com/phpstan/phpdoc-parser/blob/e352d065af1ae9b41c12d1dfd309e90f7b1f55c9/.gitattributes#L3

To be added:
- `.editorconfig`
- `.github`
- `build-abnfgen.sh`
- `build.xml`
- `phpstan.neon`
- `phpcs.xml`
- `build-cs`

Maybe also: `doc/grammars`

To be removed:
- `.travis.yml`

ondrejmirtes:
Feel free to send a PR. Closing as I don't consider this a bug, but it's a perfect contribution opportunity :)

PR #73, opened 2021-04-10
Update .gitattributes

williamdes:
Fixes: #71 

To test: `git archive HEAD --format zip --output=repo.zip`

Before: `41,0 KiB (42 028 octets)`
After: `31,7 KiB (32 488 octets)`


ondrejmirtes:
Thank you.

PR #80, opened 2021-08-30
Do not add space if empty string

alexndlm:


src/Ast/PhpDoc/PhpDocNode.php
```diff
@@ -271,7 +271,13 @@ public function getTypeAliasImportTagValues(string $tagName = '@phpstan-import-t
 
 	public function __toString(): string
 	{
-		return "/**\n * " . implode("\n * ", $this->children) . "\n */";
+		$children = array_map(
+			static function ($child): string {
+				return (string) $child === '' ? '' : ' ' . $child;
```
JanTvrdik, line 276: This will render each child twice.

src/Ast/PhpDoc/PhpDocNode.php
```diff
@@ -271,7 +271,13 @@ public function getTypeAliasImportTagValues(string $tagName = '@phpstan-import-t
...
+			},
+			$this->children
+		);
+		return "/**\n *" . implode("\n *", $children) . "\n */";
```
JanTvrdik, line 280:
```suggestion
		$children = array_map(
			static function (PhpDocChildNode $child): string {
				$s = (string) $child;
				return $s === '' ? '' : ' ' . $s;
			},
			$this->children
		);
		return "/**\n *" . implode("\n *", $children) . "\n */";
```

src/Ast/PhpDoc/PhpDocNode.php
alexndlm: I agree, thanks for the fix.

ondrejmirtes:
Thank you both!

alexndlm:
Thank you for making the release so quickly.

PR #81, opened 2021-09-05
Allow parsing class constants with more than one wildcard in the constant name

Seldaek:
Refs phpstan/phpstan#5534

### DIFF
--- a/src/Parser/ConstExprParser.php
+++ b/src/Parser/ConstExprParser.php
@@ -56,15 +56,29 @@ class ConstExprParser
 
 			if ($tokens->tryConsumeTokenType(Lexer::TOKEN_DOUBLE_COLON)) {
 				$classConstantName = '';
-				if ($tokens->currentTokenType() === Lexer::TOKEN_IDENTIFIER) {
-					$classConstantName .= $tokens->currentTokenValue();
-					$tokens->consumeTokenType(Lexer::TOKEN_IDENTIFIER);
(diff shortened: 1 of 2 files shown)


src/Parser/ConstExprParser.php
```diff
@@ -61,10 +61,18 @@ public function parse(TokenIterator $tokens, bool $trimStrings = false): Ast\Con
 					$tokens->consumeTokenType(Lexer::TOKEN_IDENTIFIER);
 					if ($tokens->tryConsumeTokenType(Lexer::TOKEN_WILDCARD)) {
 						$classConstantName .= '*';
+						if ($tokens->currentTokenType() === Lexer::TOKEN_IDENTIFIER) {
```
ondrejmirtes, line 64: I feel like this needs a `while` loop - so that we can have multiple wildcards in the name :)

ondrejmirtes (changes requested):

src/Parser/ConstExprParser.php
Seldaek: OK sure, might as well do a complete solution while I'm at it I guess.

ondrejmirtes: Yes, thank you :)

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #82, opened 2021-09-13
Fix wildcard parsing to end if a space follows

Seldaek:
Fixes phpstan/phpstan#5628

I believe this is correct.. as soon as a whitespace was skipped after a `*`, it means the `x::*` or `x::foo*` or `x::*foo*` expression ended and we should stop consuming right away.

I think it was only a problem with `@return` because there is no $var there so it is followed by a TOKEN_IDENTIFIER there, instead of TOKEN_VARIABLE in the others. Anyway the fix should be applicable to all circumstances if I am not mistaken :)

### DIFF
--- a/src/Parser/ConstExprParser.php
+++ b/src/Parser/ConstExprParser.php
@@ -70,6 +70,10 @@ class ConstExprParser
 						$classConstantName .= '*';
 						$lastType = Lexer::TOKEN_WILDCARD;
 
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you!

/cc @JanTvrdik Do you have any other suggestion? We're fixing a case where the description became part of the ConstFetchNode: `'/** @return self::* example description */'`

ondrejmirtes:
Alright, I'm gonna merge it, thank you! :)

PR #92, opened 2022-01-29
PhpDocNode: Swap array_column() and array_filter()

jlherren:
See phpstan/phpstan-src#948


ondrejmirtes:
Awesome, thank you! Please rebase your PR in phpstan-src so that we see it passes :)

PR #106, opened 2022-03-25
Support conditional types

rvanvelzen:
Part of implementation for phpstan/phpstan#3853

### DIFF
--- a/doc/grammars/type.abnf
+++ b/doc/grammars/type.abnf
@@ -3,6 +3,10 @@
 ; ---------------------------------------------------------------------------- ;
 
 Type
+	= Atomic [Union / Intersection]
+	/ Nullable
+
+ParenthesizedType
 	= Atomic [Union / Intersection / Conditional]
 	/ Nullable
 
@@ -21,7 +25,7 @@ Nullable
 Atomic
 	= TokenIdentifier [Generic / Callable / Array]
 	/ TokenThisVariable
-	/ TokenParenthesesOpen Type TokenParenthesesClose [Array]
+	/ TokenParenthesesOpen ParenthesizedType TokenParenthesesClose [Array]
 
 Generic
 	= TokenAngleBracketOpen Type *(TokenComma Type) TokenAngleBracketClose
--- a/doc/grammars/type.abnf
+++ b/doc/grammars/type.abnf
@@ -13,7 +13,7 @@ Intersection
 	= 1*(TokenIntersection Atomic)
 
 Conditional
-    = 1*ByteHorizontalWs "is" [1*ByteHorizontalWs "not"] 1*ByteHorizontalWs Atomic TokenNullable Atomic TokenColon Atomic
+    = 1*ByteHorizontalWs TokenIs [TokenNot] Atomic TokenNullable Atomic TokenColon Atomic
 
 Nullable
 	= TokenNullable TokenIdentifier [Generic]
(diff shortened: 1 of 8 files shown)


src/Ast/Type/ConditionalTypeNode.php
```diff
@@ -0,0 +1,49 @@
...
+	use NodeAttributes;
+
+	/** @var TypeNode */
+	public $subjectType;
+
+	/** @var TypeNode */
+	public $targetType;
+
+	/** @var TypeNode */
+	public $trueType;
```
ondrejmirtes: I'd rather have `if` type and `else` type, like here: https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Expr/Ternary.php

src/Parser/TokenIterator.php
```diff
@@ -59,7 +60,7 @@ public function currentTokenOffset(): int
 
 	public function isCurrentTokenValue(string $tokenValue): bool
 	{
-		return $this->tokens[$this->index][Lexer::VALUE_OFFSET] === $tokenValue;
+		return strcasecmp($this->tokens[$this->index][Lexer::VALUE_OFFSET], $tokenValue) === 0;
```
ondrejmirtes: What is this change for?

src/Parser/TypeParser.php
```diff
@@ -44,7 +46,9 @@ public function parse(TokenIterator $tokens): Ast\Type\TypeNode
 	private function parseAtomic(TokenIterator $tokens): Ast\Type\TypeNode
 	{
 		if ($tokens->tryConsumeTokenType(Lexer::TOKEN_OPEN_PARENTHESES)) {
+			$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
```
ondrejmirtes: This surely has some repercussions for some other parsed types. Can you add some tests demonstrating that?

ondrejmirtes (changes requested):
Also please check this article: https://psalm.dev/articles/conditional-love

From what I understand, you implemented this syntax in this PR:

```php
   * @psalm-return (
   *     T is int
   *     ? static
   *     : array<static>
   * ) 
```

I'd also like to support this:

```php
   * @psalm-return (
   *     $id is int
   *     ? static
   *     : array<static>
   * ) 
```

Since `$id` is not a type, feel free to do it as a completely different AST node.

Thanks! This looks promising.

src/Parser/TokenIterator.php
rvanvelzen: This supports e.g. `T IS NOT NULL`. If you prefer only to support lowercase (i.e. `T is not null`) I'll revert this

ondrejmirtes: Yes, please revert, I'm fine with supporting only a single variant :)

src/Parser/TypeParser.php
rvanvelzen: Sure. I could also provide a separate PR for this change?

rvanvelzen:
> I'd also like to support this:
> 
> ```
>    * @psalm-return (
>    *     $id is int
>    *     ? static
>    *     : array<static>
>    * ) 
> ```


I was planning to implement that syntax separately in a follow-up PR, but I'm fine with including it here as well.

> Since `$id` is not a type, feel free to do it as a completely different AST node.

Do you mean a separate `ConditionalTypeForVariable` node with a plain `$subjectName` property, or have `$subject` in `ConditionalType` be a union of `TypeNode|ParameterReferenceNode`? (All names up for debate).

ondrejmirtes:
Separate `ConditionalTypeForParameter` with a plain string `$parameterName` property :) Thank you very much.

rvanvelzen:
I believe I addresses everything. Please let me know if more changes are necessary

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

doc/grammars/type.abnf
```diff
@@ -6,19 +6,26 @@ Type
...
+	/ Nullable
+
 Union
 	= 1*(TokenUnion Atomic)
 
 Intersection
 	= 1*(TokenIntersection Atomic)
 
+Conditional
+    = 1*ByteHorizontalWs TokenIs [TokenNot] Atomic TokenNullable Atomic TokenColon Atomic
```
staabm, line 20: not sure how/wheter these `abnf` file ares sensitive to whitespace, but the newly introduced lines use spaces, not tabs

PR #118, opened 2022-04-22
Extend allowed types for conditional for parameter

rvanvelzen:
Conditional allows more whitespace between operators and union types, but this was omitted for conditional for parameter


ondrejmirtes:
Thank you!

PR #119, opened 2022-04-28
Implement offset access type syntax

rvanvelzen:
This implements offset access types keeping as much BC as possible. Notably, whitespace between the type and `[` is not allowed (so `A[B]` is valid but `A [B]` parses as just `A`, as before)

### DIFF
--- a/src/Ast/Type/OffsetAccessTypeNode.php
+++ b/src/Ast/Type/OffsetAccessTypeNode.php
@@ -1,0 +1,29 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast\Type;
+
+use PHPStan\PhpDocParser\Ast\NodeAttributes;
+
+class OffsetAccessTypeNode implements TypeNode
+{
+
+	use NodeAttributes;
+
+	/** @var TypeNode */
+	public $type;
+
+	/** @var TypeNode */
+	public $offset;
+
+	public function __construct(TypeNode $type, TypeNode $offset)
+	{
+		$this->type = $type;
+		$this->offset = $offset;
+	}
+
+	public function __toString(): string
+	{
+		return $this->type . '[' . $this->offset . ']';
+	}
+
+}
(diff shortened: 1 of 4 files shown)


ondrejmirtes (changes requested):
A test for the `A [B]` scenario wouldn't hurt :)

tests/PHPStan/Parser/PhpDocParserTest.php
```diff
@@ -972,6 +973,23 @@ public function provideReturnTagsData(): Iterator
 			]),
 		];
 
+		yield [
```
rvanvelzen, line 976: Tests for `A [B]` already exist - just a few lines above

rvanvelzen:
@JanTvrdik what's the reason for the "confused" reaction? Do you disagree with the idea in general, with the implementation, or something else?

ondrejmirtes:
The context: https://github.com/phpstan/phpstan/issues/7094

src/Ast/Type/OffsetAccessTypeNode.php
```diff
@@ -0,0 +1,29 @@
...
+
+	use NodeAttributes;
+
+	/** @var TypeNode */
+	public $type;
+
+	/** @var TypeNode */
+	public $offset;
+
+	public function __construct(TypeNode $type, TypeNode $offset)
```
ondrejmirtes, line 18: Does it make sense to allow any TypeNode here? Shouldn't it be narrowed down to just IdentifierTypeNode?

ondrejmirtes (changes requested):

src/Ast/Type/OffsetAccessTypeNode.php
rvanvelzen: While it might make no sense in practice, `array{foo: string}['foo']` could resolve to `string`.

ondrejmirtes: Alright, make sure to add a test case for this in phpstan-src when you work on it :) Thanks :)

PR #120, opened 2022-04-29
Add assert syntax

rvanvelzen:
Syntax for phpstan/phpstan#7110

### DIFF
--- a/tests/PHPStan/Parser/PhpDocParserTest.php
+++ b/tests/PHPStan/Parser/PhpDocParserTest.php
@@ -3565,6 +3565,35 @@ some text in the middle'
 				),
 			]),
 		];
+
+		yield [
+			'OK assert-if-true',
+			'/** @phpstan-assert-if-true Type $var */',
+			new PhpDocNode([
+				new PhpDocTagNode(
+					'@phpstan-assert-if-true',
+					new AssertTagValueNode(
+						new IdentifierTypeNode('Type'),
+						'$var',
+						''
+					)
+				),
+			]),
+		];
+		yield [
+			'OK assert-if-false',
+			'/** @phpstan-assert-if-false Type $var */',
+			new PhpDocNode([
+				new PhpDocTagNode(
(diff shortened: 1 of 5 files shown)


tests/PHPStan/Parser/PhpDocParserTest.php
```diff
@@ -3480,6 +3482,91 @@ public function provideTypeAliasImportTagsData(): Iterator
...
+						'$var',
+						''
+					)
+				),
+			]),
+		];
+
+		yield [
+			'invalid $this->method()',
+			'/** @phpstan-assert Type $this->method() */',
```
rvanvelzen, line 3556: This would be a nice future addition

rvanvelzen:
I've actually got a more-or-less working implementation with this, but I think allowing more complicated expression (e.g. `$this->getSomething()`) isn't _too_ hard and would at least be nice to support here.

ondrejmirtes (changes requested):
Yes, it'd be nice to have properties and methods too :) Thank you!

ondrejmirtes:
BTW @rvanvelzen Did you start working on the counterpart in phpstan-src? If not then I can merge this as it is, get it working in phpstan-src (I have some ideas about that), and we can do methods+properties as follow-up PRs :) What do you think?

rvanvelzen:
I did work on it a bit but feel free to get it working :)

ondrejmirtes:
Thank you!

tests/PHPStan/Parser/PhpDocParserTest.php
ZebulanStanphill: Does this mean the current implementation doesn't support class props (e.g. `$this->myProp`), or are you only talking about methods?

rvanvelzen: Any syntax other than plain variables hasn't been implemented yet. It will definitely happen in the future.

PR #127, opened 2022-06-08
Allow omitting @param type

### LINKED ISSUE #77: @param type can be omitted
According to https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/param.html, the type may be omitted. This is especially useful in PHP 8 applications, which can almost fully typehint all arguments. The type in the PHPdoc in this case is redundant and may only lead to out-of-sync types.

PHPstorm and Psalm already appear to support this. However, this is not yet the case with PHPstan: https://phpstan.org/r/691a2424-e087-45a0-a9ce-774a57f887ec

(fyi, we're about to remove the types in Symfony, which is why I checked support in the popular static analysis libraries)

rvanvelzen:
Fixes #77

### DIFF
--- a/tests/PHPStan/Parser/PhpDocParserTest.php
+++ b/tests/PHPStan/Parser/PhpDocParserTest.php
@@ -27,6 +27,7 @@ use PHPStan\PhpDocParser\Ast\PhpDoc\TemplateTagValueNode;
 use PHPStan\PhpDocParser\Ast\PhpDoc\ThrowsTagValueNode;
 use PHPStan\PhpDocParser\Ast\PhpDoc\TypeAliasImportTagValueNode;
 use PHPStan\PhpDocParser\Ast\PhpDoc\TypeAliasTagValueNode;
+use PHPStan\PhpDocParser\Ast\PhpDoc\TypelessParamTagValueNode;
 use PHPStan\PhpDocParser\Ast\PhpDoc\UsesTagValueNode;
 use PHPStan\PhpDocParser\Ast\PhpDoc\VarTagValueNode;
 use PHPStan\PhpDocParser\Ast\Type\ArrayTypeNode;
@@ -64,6 +65,7 @@ class PhpDocParserTest extends TestCase
 	/**
 	 * @dataProvider provideTagsWithNumbers
 	 * @dataProvider provideParamTagsData
+	 * @dataProvider provideTypelessParamTagsData
 	 * @dataProvider provideVarTagsData
 	 * @dataProvider provideReturnTagsData
 	 * @dataProvider provideThrowsTagsData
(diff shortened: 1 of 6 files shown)


ondrejmirtes:
We should be able to do this without breaking BC. Because `@param` without a type is essentially useless, I feel like we shouldn't mix it with other useful `@param` tags that have a type.

So my suggestion is:

1) Do not make `$type` in `src/Ast/PhpDoc/ParamTagValueNode.php` nullable.
2) Create a separate `TypelessParamTagValueNode` that can still carry `$description`.
3) `PhpDocNode::getParamTagValues()` WIL NOT return `TypelessParamTagValueNode`, it needs to be filtered out.
4) Add `PhpDocNode::getTypelessParamTagValues()`.
5) I'll release that in minor phpdoc-parser version.
6) Add some tests to phpstan-src that demonstrate this is no longer a parse error, and potentially get and make use of some data carried by `getTypelessParamTagValues()` (but I don't think we use parameter description anywhere).

Thank you!

tests/PHPStan/Parser/PhpDocParserTest.php
```diff
@@ -395,6 +397,89 @@ public function provideParamTagsData(): Iterator
...
+						'$foo',
+						'description',
+						true
+					)
+				),
+			]),
+		];
+
+		yield [
+			'invalid without type and description',
```
ondrejmirtes: I think the conclusion of https://github.com/phpstan/phpdoc-parser/issues/77 is that `/** @param $foo */` shouldn't be a parse error either. The description can be an empty string, and if someone wants, they can have a rule to mark such tag as useless, but it would no longer be an objective bug to have such PHPDoc in the code...

So the grammar description + parser should be update to allow this tag.

ondrejmirtes (changes requested):

ondrejmirtes:
Perfect, thank you!

wouterj:
Thank you Ondřej and Richard!

Wirone:
This one made our baseline files thinner by ~3k entries 🙈 I am not 100% convinced that this is good behaviour but I understand the reasoning behind it. Cheers! 🍻

p-golovin:
Is it possible to revert the old behavior by configuration flag?

rvanvelzen:
Why?

p-golovin:
It was useful for us to detect this errors because they are usually show old code with bad type hinting (in our code base). We check and fix them step by step. So we need to stay at 1.7.11 version before will fix them all.

ondrejmirtes:
@p-golovin As said earlier in this thread, you'll get the same errors on level 6 (missing parameter typehint).

p-golovin:
@ondrejmirtes Thank you. We'll try to raise level to 6 (now we make fixes on level 3).

Wirone:
> As said earlier in this thread, you'll get the same errors on level 6 (missing parameter typehint).

Really? We're on level 6 and upgrading caused many "ignored ... not found", so we had to regenerate baselines and maaaaaaany ignores were removed. 🤔

EDIT: remember, do not comment when too tired to understand 😉

wouterj:
see https://github.com/phpstan/phpdoc-parser/issues/77#issuecomment-1156370603

PR #128, opened 2022-06-08
Require whitespace before description with limited start tokens

### LINKED ISSUE #125: Unnamed vargs type in callable/Closure phpdoc is silently ignored
# Bug report

Currently

```
@return \Closure(int ...$u, string): string
```

phpdoc works correctly, but

```
@return \Closure(...int, object): string
```

is silently ignored. No wrong phpdoc error, no error when this type does not match the type of the returned value.

Even if such type is not documented as supported - https://phpstan.org/writing-php-code/phpdoc-types#callables - I belive it should be supported as standard (non-vargs) unnamed parameters are supported.

https://phpstan.org/r/f3779e03-f24f-43e7-8d73-b27973bbd15c

### Expected output

The 2nd phpdoc must emit a phpstan error, the type (return type of the `Closure`) does not match the result.

rvanvelzen:
Fixes #125

### DIFF
--- a/tests/PHPStan/Parser/PhpDocParserTest.php
+++ b/tests/PHPStan/Parser/PhpDocParserTest.php
@@ -1310,6 +1310,58 @@ class PhpDocParserTest extends TestCase
 				),
 			]),
 		];
+
+		yield [
+			'OK variadic callable',
+			'/** @return \Closure(int ...$u, string): string */',
+			new PhpDocNode([
+				new PhpDocTagNode(
+					'@return',
+					new ReturnTagValueNode(
+						new CallableTypeNode(
+							new IdentifierTypeNode('\Closure'),
+							[
+								new CallableTypeParameterNode(
+									new IdentifierTypeNode('int'),
+									false,
+									true,
+									'$u',
+									false
+								),
+								new CallableTypeParameterNode(
+									new IdentifierTypeNode('string'),
+									false,
+									false,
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
I really like this! I'm gonna check the integration tests in PHPStan before tagging a release here because there might be some real-world impact.

ondrejmirtes:
Thank you!

tests/PHPStan/Parser/PhpDocParserTest.php
```diff
@@ -1310,6 +1310,58 @@ public function provideReturnTagsData(): Iterator
...
+		];
+
+		yield [
+			'invalid variadic callable',
+			'/** @return \Closure(...int, string): string */',
+			new PhpDocNode([
+				new PhpDocTagNode(
+					'@return',
+					new InvalidTagValueNode(
+						'\Closure(...int, string): string',
```
mvorisek, line 1354: thank you for throwing at least

as all other types in `\Closure` phpdoc does not require a variable name, can support for unnamed variadic phpdoc be added?

rvanvelzen: It is already supported, but the type always has to come first.

mvorisek: I didn't know, thank you ❤️

PR #129, opened 2022-06-08
Support all atomic types as nullable types

### LINKED ISSUE #124: Type parsing when combinding nullable and consts-types
# Bug report

Using nullable `?` on Consts-types isn't parsed as expected

https://phpstan.org/r/d9e3f9e8-eca3-4380-8dda-ecf015d6a322

image

### Expected output

No error, expected type `?Currency::CURRENCY_*` and `null|Currency::CURRENCY_*` to be equal.

### Did PHPStan help you today? Did it make you happy in any way?

Happy as in "Yes, catched that before commitin/pushing/deploying it"

rvanvelzen:
Fixes #124

### DIFF
--- a/doc/grammars/type.abnf
+++ b/doc/grammars/type.abnf
@@ -20,7 +20,7 @@ Conditional
     = 1*ByteHorizontalWs TokenIs [TokenNot] Atomic TokenNullable Atomic TokenColon Atomic
 
 Nullable
-	= TokenNullable TokenIdentifier [Generic]
+	= TokenNullable Atomic
 
 Atomic
 	= TokenIdentifier [Generic / Callable / Array]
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Perfect! I'm eager to merge this too :)

ondrejmirtes:
Thank you!

ondrejmirtes:
FYI I'm gonna try phpdoc-parser in phpstan-src again after this suggestion https://github.com/phpstan/phpstan-src/pull/1406#issuecomment-1150059830 - I think it's really necessary to make it optional and non-default because the new behaviour is much stricter and will make a lot of projects fail, so it's a good fit for bleedingEdge (and will allow us to release it in a minor phpdoc-parser version).

ondrejmirtes:
But I understand it's not trivial. Ideally I'd like to see all current data providers to be tested with both `true` / `false` and only those that differ to be tested in a different test case method probably.

rvanvelzen:
I'll hopefully get to it tomorrow :)

PR #131, opened 2022-06-09
Make requiring whitespace before description optional

rvanvelzen:
Per https://github.com/phpstan/phpstan-src/pull/1406#issuecomment-1150059830


ondrejmirtes:
Perfect, thank you!

PR #133, opened 2022-06-10
Optimize Lexer::tokenize()

rvanvelzen:
Parsing the docblock from [CarbonInterface](https://github.com/briannesbitt/Carbon/blob/4c4f99e79a9ddea9bb1458a4cca3d4c06c24beff/src/Carbon/CarbonInterface.php) 50 times takes ~3.5s before this change, and ~1s after.


ondrejmirtes:
Thank you!

PR #134, opened 2022-06-10
Optimize generated regex

rvanvelzen:
Matching the most common elements first provides another substantial boost.


ondrejmirtes:
Wow, thank you :)

PR #139, opened 2022-06-20
Support properties/methods in assert annotations

rvanvelzen:


tests/PHPStan/Parser/PhpDocParserTest.php
```diff
@@ -3797,24 +3797,88 @@ public function provideAssertTagsData(): Iterator
...
 
 		yield [
-			'invalid $this->method()',
-			'/** @phpstan-assert Type $this->method() */',
+			'OK $var->method()',
+			'/** @phpstan-assert Type $var->method() */',
+			new PhpDocNode([
+				new PhpDocTagNode(
+					'@phpstan-assert',
+					new AssertTagValueNode(
```
ondrejmirtes: Would be much nicer to have `AssertTagValueNode` only for parameters, and have separate nodes for properties and methods :) Like `AssertTagMethodValueNode` + `AssertTagPropertyValueNode`.

I imagine these two nodes would still have `public string $parameter` to represent the `@var`, but additionally they would also have `public string $method` and `public string $property` respectively :)

Yeah, unfortunately we can't change the `AssertTagValueNode` name to `AssertTagParameterValueNode` now, we're stuck with it to 2.0, but it's just a minor detail.

ondrejmirtes (changes requested):

tests/PHPStan/Parser/PhpDocParserTest.php
rvanvelzen: I've taken a slightly different approach that keeps the parser code a bit simpler, but I now realize that is a BC break as well :disappointed: 

I'll update it later today

rvanvelzen:
I'm not 100% satisfied with the implementation, but it gets the job done

ondrejmirtes (approved):

ondrejmirtes:
Thank you! :)

PR #140, opened 2022-06-26
Support @template-contravariant tags

autaut03:
This allows implementing https://github.com/phpstan/phpstan/issues/7502 on PHPStan side.

### DIFF
--- a/src/Parser/PhpDocParser.php
+++ b/src/Parser/PhpDocParser.php
@@ -178,6 +178,9 @@ class PhpDocParser
 				case '@template-covariant':
 				case '@phpstan-template-covariant':
 				case '@psalm-template-covariant':
+				case '@template-contravariant':
+				case '@phpstan-template-contravariant':
+				case '@psalm-template-contravariant':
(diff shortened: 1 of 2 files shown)


src/Parser/PhpDocParser.php
```diff
@@ -178,6 +178,9 @@ public function parseTagValue(TokenIterator $tokens, string $tag): Ast\PhpDoc\Ph
 				case '@template-covariant':
 				case '@phpstan-template-covariant':
 				case '@psalm-template-covariant':
+				case '@template-contravariant':
+				case '@phpstan-template-contravariant':
+				case '@psalm-template-contravariant':
```
autaut03, line 183: Psalm doesn't currently support @template-contravariant tags either, so I'm unsure whether this should be included.

Support [would be welcomed](https://github.com/vimeo/psalm/issues/5252), so I'd rather keep this.

ondrejmirtes: Yes, there's no harm in having `@psalm-template-contravariant` here :)

ondrejmirtes:
Thank you!

PR #141, opened 2022-07-20
Added ability to parse object shapes.

mad-briller:
partially fixes https://github.com/phpstan/phpstan/issues/2923

### DIFF
--- a/doc/grammars/type.abnf
+++ b/doc/grammars/type.abnf
@@ -73,7 +73,7 @@ ShapeItem
     / Type
 
 ObjectShape
-	= Shape
+    = Shape
 
 ObjectShapeItem
     = ShapeItem
--- a/doc/grammars/type.abnf
+++ b/doc/grammars/type.abnf
@@ -60,11 +60,23 @@ Array
 	= 1*(TokenSquareBracketOpen TokenSquareBracketClose)
 
 ArrayShape
-    = TokenCurlyBracketOpen ArrayShapeItem *(TokenComma ArrayShapeItem) TokenCurlyBracketClose
+	= Shape
 
 ArrayShapeItem
+    = ShapeItem
+
+Shape
+    = TokenCurlyBracketOpen ShapeItem *(TokenComma ShapeItem) TokenCurlyBracketClose
+
+ShapeItem
     = (ConstantString / ConstantInt / TokenIdentifier) TokenNullable TokenColon Type
     / Type
+
+ObjectShape
+	= Shape
+
+ObjectShapeItem
+    = ShapeItem
 
 ; ---------------------------------------------------------------------------- ;
 ;  ConstantExpr                                                                ;
--- a/doc/grammars/type.abnf
+++ b/doc/grammars/type.abnf
@@ -63,20 +63,20 @@ ArrayShape
 	= Shape
 
 ArrayShapeItem
-    = ShapeItem
+	= ShapeItem
 
 Shape
-    = TokenCurlyBracketOpen ShapeItem *(TokenComma ShapeItem) TokenCurlyBracketClose
+	= TokenCurlyBracketOpen ShapeItem *(TokenComma ShapeItem) TokenCurlyBracketClose
 
 ShapeItem
-    = (ConstantString / ConstantInt / TokenIdentifier) TokenNullable TokenColon Type
-    / Type
+	= (ConstantString / ConstantInt / TokenIdentifier) TokenNullable TokenColon Type
+	/ Type
 
 ObjectShape
-    = Shape
+	= Shape
 
 ObjectShapeItem
-    = ShapeItem
+	= ShapeItem
 
 ; ---------------------------------------------------------------------------- ;
 ;  ConstantExpr                                                                ;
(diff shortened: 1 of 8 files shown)


doc/grammars/type.abnf
```diff
@@ -60,12 +60,24 @@ Array
...
+ShapeItem
     = (ConstantString / ConstantInt / TokenIdentifier) TokenNullable TokenColon Type
     / Type
 
+ObjectShape
+	= Shape
+
+ObjectShapeItem
+    = ShapeItem
+
```
mad-briller, line 80: this is my first time encountering ABNF, so not super confident on this

src/Parser/TypeParser.php
```diff
@@ -468,53 +470,76 @@ private function tryParseArrayOrOffsetAccess(TokenIterator $tokens, Ast\Type\Typ
...
 		try {
 			$tokens->pushSavePoint();
-			$key = $this->parseArrayShapeKey($tokens);
+			$key = $this->parseShapeKey($tokens);
 			$optional = $tokens->tryConsumeTokenType(Lexer::TOKEN_NULLABLE);
 			$tokens->consumeTokenType(Lexer::TOKEN_COLON);
 			$value = $this->parse($tokens);
 			$tokens->dropSavePoint();
 
+			if ($type->name === 'object') {
```
mad-briller, line 530: not quite sure on this approach to re-using the `parseArrayShapeItem` code, it seems fairly brittle but the other option would be to `return [$key, $optional, $value];` and then construct the correct value in `parseShape`, but that felt pretty messy too.

doc/grammars/type.abnf
```diff
@@ -60,12 +60,24 @@ Array
...
+Shape
+    = TokenCurlyBracketOpen ShapeItem *(TokenComma ShapeItem) TokenCurlyBracketClose
+
+ShapeItem
     = (ConstantString / ConstantInt / TokenIdentifier) TokenNullable TokenColon Type
     / Type
 
+ObjectShape
+	= Shape
```
staabm, line 76: nit: tabs vs spaces

mad-briller: good spot

the `.editorconfig` doesn't contain an entry for `.abnf` files so i'll add that aswell

JanTvrdik:
I don't think it makes any sense to allow optinal keys for object shapes (because object tuples are nonsense), so you likely cannot directly reuse array shape parser and/or grammar.

Edit: it also probably does not make sense to allow constant strings / constant integers as object shape keys. probably only identifiers make sense.

doc/grammars/type.abnf
```diff
@@ -60,11 +60,23 @@ Array
...
-    = (ConstantString / ConstantInt / TokenIdentifier) TokenNullable TokenColon Type
-    / Type
+	= ShapeItem
+
+Shape
+	= TokenCurlyBracketOpen ShapeItem *(TokenComma ShapeItem) TokenCurlyBracketClose
+
+ShapeItem
+	= (ConstantString / ConstantInt / TokenIdentifier) TokenNullable TokenColon Type
+	/ Type
```
JanTvrdik, line 73: As mentioned in the comment, this only makes sense for array shapes. For object shapes you need sth like


~~~abnf
ObjectShapeItem = TokenIdentifier [TokenNullable] TokenColon Type
~~~

mad-briller: thanks for the feedback

silly oversight there with allowing for `object{string, string}`, i'll rework this tomorrow if i can

src/Parser/TypeParser.php
mad-briller: reusing the parser is not a great idea due to issues raised in https://github.com/phpstan/phpdoc-parser/pull/141#issuecomment-1190809481

going to rework the approach

ondrejmirtes:
Hi, I appreciate this, but before merging this, I'd like to see some work in progress on phpstan-src side too. To make sure there isn't a 6 months (or an infinite) gap between the progress in phpdoc-parser and phpstan-src :) So please try to implement `ObjectShapeType` in phpstan-src and change the phpdoc-parser dependency in composer.json to point at your repo and branch.

The `ObjectShapeType` we'll have to implement is going to be a combination of ObjectWithoutClassType and ConstantArrayType (https://apiref.phpstan.org/1.8.x/PHPStan.Type.ObjectWithoutClassType.html, https://apiref.phpstan.org/1.8.x/PHPStan.Type.Constant.ConstantArrayType.html).

The hardest part about adding a new type is that it has to play well with all the current types. It has to correctly implement all the Type methods, but especially Type::accepts and Type::isSuperTypeOf().

`Type::accepts()` can have some basic tests in a rule that checks type acceptance, like CallMethodsRuleTest. And of course you can add ObjectShapeTypeTest::testAccepts() to comfortably test a lot of combinations.

`Type::isSuperTypeOf()` is about this https://phpstan.org/developing-extensions/type-system#querying-a-specific-type and is best tested in TypeCombinatorTest::dataUnion() and TypeCombinatorTest::dataIntersect(). We'll have to answer some difficult questions like which type is the supertype of which another type :)

BTW: About a different issue - implementing the "generic callable" syntax here and in phpstan-src is definitely an easier task so feel free to start with it :)

mad-briller:
thank you for the feedback ondrej.

i started with this as it seemed like it would be the easier thing to add to phpdoc-parser to get more familiar with it, to then allow me to do the generic callable syntax! but i guess i was wrong :D

dimitriskr:
The error of the check is irrelevant to the work of the PR. It is about allowing composer plugin to execute
Is it possible to re-run the checks?

mad-briller:
superceded by https://github.com/phpstan/phpdoc-parser/releases/tag/1.18.0

ondrejmirtes:
Hey, it'd be great if you could update the grammar, I totally forgot about that!

PR #143, opened 2022-08-09
Unescape constant strings

### LINKED ISSUE #142: Cannot use backslash as a literal string in conditional return type
i can't get phpstan to recognise `\\` in a literal string:
https://phpstan.org/r/41f1530f-eaee-4e7d-8bfc-e7ccc7cfc7a7

using any other sting there makes it work just fine:
https://phpstan.org/r/a884b37d-682b-45a4-b891-1725ffd1c428

_Originally posted by @mad-briller in https://github.com/phpstan/phpstan/issues/7483#issuecomment-1200168302_

rvanvelzen:
Fix #142

### DIFF
--- a/src/Parser/ConstExprParser.php
+++ b/src/Parser/ConstExprParser.php
@@ -4,11 +4,26 @@ namespace PHPStan\PhpDocParser\Parser;
 
 use PHPStan\PhpDocParser\Ast;
 use PHPStan\PhpDocParser\Lexer\Lexer;
+use function chr;
+use function hexdec;
+use function octdec;
+use function preg_replace_callback;
+use function str_replace;
 use function strtolower;
-use function trim;
+use function substr;
 
 class ConstExprParser
 {
+
+	private const REPLACEMENTS = [
+		'\\' => '\\',
+		'n' => "\n",
+		'r' => "\r",
+		't' => "\t",
+		'f' => "\f",
+		'v' => "\v",
+		'e' => "\x1B",
+	];
 
 	public function parse(TokenIterator $tokens, bool $trimStrings = false): Ast\ConstExpr\ConstExprNode
 	{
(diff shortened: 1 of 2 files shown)


JanTvrdik:
I would expect double quoted string to support the same escaping sequences as PHP.
See https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Scalar/String_.php#L71-L130

rvanvelzen:
I initially was going to do so, but didn't for [parity with Psalm](https://github.com/vimeo/psalm/blob/36d5a2a83c1d7b3e2aca1cfe9e9dcc916619e962/src/Psalm/Internal/Type/TypeTokenizer.php#L166). I'd be more than happy to mirror PHP's functionality though, if @ondrejmirtes also agrees :)

ondrejmirtes:
Yes, parity with PHP would be great, we're in a very edge case territory anyway. And inform the Psalm maintainers afterwards 😊 BTW is this a BC break? Do we need a feature flag for the new behaviour?

src/Parser/ConstExprParser.php
```diff
@@ -137,4 +130,17 @@ private function parseArrayItem(TokenIterator $tokens): Ast\ConstExpr\ConstExprA
 		return new Ast\ConstExpr\ConstExprArrayItemNode($key, $value);
 	}
 
+	private function trimString(string $string): string
```
JanTvrdik: decode or parse string would be better name

rvanvelzen:
I've basically copied most of the implementation from PHP-Parser, but that probably needs to be mentioned somewhere. Any suggestions?

ondrejmirtes:
A github.com permalink in a code comment would be fine I think.

ondrejmirtes:
This is gonna need some modification and a bleedingEdge toggle in phpstan-src too, right?

rvanvelzen:
It should only require a feature toggle :)

ondrejmirtes:
What about the `bool $trimStrings = false`? parameter? Shouldn't we eventually get rid of it? In PHPStan it's probably called with false...

src/Parser/ConstExprParser.php
```diff
@@ -137,4 +156,69 @@ private function parseArrayItem(TokenIterator $tokens): Ast\ConstExpr\ConstExprA
...
+					return self::codePointToUtf8(hexdec($matches[2]));
+				}
+
+				return chr(octdec($str));
+			},
+			$str
+		);
+	}
+
+	private static function codePointToUtf8(int $num): string
```
JanTvrdik, line 205: Would it be possible to use `mb_chr()` or `IntlChar::chr()` instead?

rvanvelzen: Neither mbstring nor intl are declared as a dependency, so that would need to be conditional. That would only make the code more complicated for little gain.

rvanvelzen:
> What about the `bool $trimStrings = false`? parameter? Shouldn't we eventually get rid of it? In PHPStan it's probably called with false...

I would definitely like to get rid of it, especially because it is used with both true/false internally. `@method` parameter default values are parsed without unescaping, which eventually leads to unescaped values in PHPStan as well. Those never really propagate anywhere though…

There isn't a pure BC-friendly approach to removing it I'm afraid, but I really don't think it should be used at all so maybe it's not that bad?

ondrejmirtes:
Thank you!

PR #146, opened 2022-09-04
Allow backslash in PHPDoc tag

kubawerlos:
As discussed in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/6619#issuecomment-1236329281.

### DIFF
--- a/src/Lexer/Lexer.php
+++ b/src/Lexer/Lexer.php
@@ -146,7 +146,7 @@ class Lexer
 
 			self::TOKEN_OPEN_PHPDOC => '/\\*\\*(?=\\s)\\x20?+',
 			self::TOKEN_CLOSE_PHPDOC => '\\*/',
-			self::TOKEN_PHPDOC_TAG => '@[a-z][a-z0-9-]*+',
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Can you please try what's it gonna do with more complicated ORM tags with parentheses and attributes inside?

ondrejmirtes:
It's nice it's not a parse error, but some dedicated node so that this can be read comfortably would be nice, right?

Do you see yourself contributing this, or should I merge the PR as it is?

Thank you.

kubawerlos:
I've stuck on `PhpDocParser::parseTagValue` - a new node would have to be added on that `switch`, right?

Also, I'm not sure if the node should be only for parenthesis and the content inside, or should unpack the content into an array (that probably should handle both `=` and `:` Doctrine's notations, right?).

So I'd like to pass on that, as it might take much longer for you to review and explain to me, that you were doing it yourself :)

ondrejmirtes:
Yeah, this is good enough as it is. Let's address that need when it comes up. I'm gonna release this as a next minor version. Thank you.

PR #147, opened 2022-09-06
Improve abnf readability

rvanvelzen:
Per https://github.com/phpstan/phpdoc-parser/commit/445ccb723cee1fd1215eb96730e0d121b60204db#r81813906

### DIFF
--- a/doc/grammars/type.abnf
+++ b/doc/grammars/type.abnf
@@ -180,13 +180,13 @@ TokenDoubleColon
 	= "::" *ByteHorizontalWs
 
 TokenThisVariable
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you! Do you think it's viable to generate the parser from the grammar?

1) To make sure the grammar is in sync
2) To not have to write the parser by hand 😊

rvanvelzen:
For just the type parser it might be doable, but it would probably become a full rewrite of the parser :smiling_face_with_tear:

ondrejmirtes:
We don't need it, it was just a question 😊

PR #148, opened 2022-09-06
(RFC) Implement template default syntax

rvanvelzen:
Proof-of-concept implementation for template default syntax (for phpstan/phpstan#4801)

### DIFF
--- a/src/Ast/PhpDoc/TemplateTagValueNode.php
+++ b/src/Ast/PhpDoc/TemplateTagValueNode.php
@@ -17,13 +17,17 @@ class TemplateTagValueNode implements PhpDocTagValueNode
 	/** @var TypeNode|null */
 	public $bound;
 
+	/** @var TypeNode|null */
+	public $default;
+
 	/** @var string (may be empty) */
 	public $description;
 
-	public function __construct(string $name, ?TypeNode $bound, string $description)
+	public function __construct(string $name, ?TypeNode $bound, string $description, ?TypeNode $default = null)
 	{
 		$this->name = $name;
 		$this->bound = $bound;
+		$this->default = $default;
 		$this->description = $description;
 	}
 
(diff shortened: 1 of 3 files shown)


orklah:
Psalm doesn't have this but it could be interesting (especially because Psalm sometimes struggle to assign values to classes with templates that were never properly created)

I doubt I'll be in position to implement this myself anytime soon though

ondrejmirtes:
Alright, I'm fine with going ahead with this. But what the syntax should be? The issue proposes a separate `@template-default` tag. What's the precedent from other languages that have this feature?

How would the syntax proposed here for `@template` work when the bound is involved?

rvanvelzen:
I've taken the syntax from [TypeScript](https://github.com/microsoft/TypeScript/pull/13487). At least [C++](https://en.cppreference.com/w/cpp/language/template_parameters) also uses this syntax, but I haven't looked at other languages.

Because the default value is tightly coupled to the type parameter (just like the bound is), I don't think it's a good idea to have a separate tag for it.

ondrejmirtes:
Nice, it makes sense to me. Could you also please submit a proof of concept to phpstan-src to make sure these are not released with a big time gap between them? :) Thank you.

What we need to think about is modifying the rules here https://github.com/phpstan/phpstan-src/tree/1.8.x/src/Rules/Generics for this new feature:

* We need to allow to omit the optional template type around here https://github.com/phpstan/phpstan-src/blob/fb08fb530f60d3510f2864b9935d2f75b29d7840/src/Rules/Generics/GenericObjectTypeCheck.php#L60-L78
* We need to check that the default type is a subtype of the bound in this class https://github.com/phpstan/phpstan-src/blob/1.8.x/src/Rules/Generics/TemplateTypeCheck.php
* We need to check that an optional template type is not followed by a non-optional one, also probably in  TemplateTypeCheck

ondrejmirtes:
Thank you!

alexander-schranz:
Nice to see this merged. 🎉 🚀

PR #154, opened 2022-10-20
Remove $trimStrings from ConstExprParser

rvanvelzen:
Per https://github.com/phpstan/phpdoc-parser/pull/143#issuecomment-1281099895

### DIFF
--- a/src/Parser/ConstExprParser.php
+++ b/src/Parser/ConstExprParser.php
@@ -33,7 +33,7 @@ class ConstExprParser
 		$this->unescapeStrings = $unescapeStrings;
 	}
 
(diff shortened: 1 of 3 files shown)


rvanvelzen:
This should be fine for the next major as it is a BC break.

ondrejmirtes:
I think we could decouple phpdoc-parser version from PHPStan so we could release 2.0 now and update to it in PHPStan 1.x :) On the other hand, this would be a PHPStan-userland BC break too, so it might not be a good idea...

ondrejmirtes:
Just checked in PHPStan and for PHPStan users it's not really a BC break because it's already called with `true`. The only exception are `@method` tags and that's fine...

ondrejmirtes:
Already done in 2.0.x.

PR #156, opened 2022-10-21
Add equality assert syntax

rvanvelzen:
As supported in Psalm: https://psalm.dev/docs/annotating_code/assertion_syntax/#equality-assertions

### DIFF
--- a/src/Ast/PhpDoc/AssertTagMethodValueNode.php
+++ b/src/Ast/PhpDoc/AssertTagMethodValueNode.php
@@ -23,15 +23,19 @@ class AssertTagMethodValueNode implements PhpDocTagValueNode
 	/** @var bool */
 	public $isNegated;
 
+	/** @var bool */
+	public $isEquality;
+
 	/** @var string (may be empty) */
 	public $description;
 
-	public function __construct(TypeNode $type, string $parameter, string $method, bool $isNegated, string $description)
+	public function __construct(TypeNode $type, string $parameter, string $method, bool $isNegated, string $description, bool $isEquality = false)
 	{
 		$this->type = $type;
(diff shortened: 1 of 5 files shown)


src/Ast/PhpDoc/AssertTagMethodValueNode.php
```diff
@@ -23,23 +23,28 @@ class AssertTagMethodValueNode implements PhpDocTagValueNode
...
 		$this->description = $description;
 	}
 
 
 	public function __toString(): string
 	{
 		$isNegated = $this->isNegated ? '!' : '';
-		return trim("{$isNegated}{$this->type} {$this->parameter}->{$this->method}() {$this->description}");
+		$isEquality = $this->isEquality ? '=' : '';
+		return trim("{$isNegated}{$isEquality}{$this->type} {$this->parameter}->{$this->method}() {$this->description}");
```
staabm, line 47: seeing this, I am wondering whether paranthesis as I used them in

https://github.com/phpstan/phpstan-src/pull/1880/files#diff-8f0c25b196228aa9450cbd3c1d8c83d6d5f33c9fe0aa1b07effe32d41275dd3eR41

are a valid thing?

rvanvelzen: They're valid since you can just parenthesize any type, but it's not necessary.

staabm: additionally, maybe its worth having a test-case for more complex ones like e.g. `@phpstan-assert-if-false !=(float|int|numeric-string) $value`

ondrejmirtes:
Thank you!

PR #160, opened 2022-11-01
Add generic support to `@method` definitions

tyabu12:
This allows implementing https://github.com/phpstan/phpstan/issues/6371 to add support`@method` generics.


src/Ast/PhpDoc/MethodTagValueGenericNode.php
```diff
@@ -0,0 +1,33 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast\PhpDoc;
+
+use PHPStan\PhpDocParser\Ast\Node;
+use PHPStan\PhpDocParser\Ast\NodeAttributes;
+use PHPStan\PhpDocParser\Ast\Type\TypeNode;
+
+class MethodTagValueGenericNode implements Node
```
ondrejmirtes: I feel like this doesn't need a new node, but TemplateTagValueNode can be reused instead. We'd also gain default template types (https://github.com/phpstan/phpdoc-parser/pull/148) automatically :)

src/Ast/PhpDoc/MethodTagValueNode.php
```diff
@@ -20,17 +20,21 @@ class MethodTagValueNode implements PhpDocTagValueNode
 	/** @var string */
 	public $methodName;
 
+	/** @var MethodTagValueGenericNode[] */
+	public $generics;
```
ondrejmirtes: This thing should be called `$templateTypes` probably.

src/Ast/PhpDoc/MethodTagValueNode.php
```diff
@@ -20,17 +20,21 @@ class MethodTagValueNode implements PhpDocTagValueNode
...
+
 	/** @var MethodTagValueParameterNode[] */
 	public $parameters;
 
 	/** @var string (may be empty) */
 	public $description;
 
-	public function __construct(bool $isStatic, ?TypeNode $returnType, string $methodName, array $parameters, string $description)
+	public function __construct(bool $isStatic, ?TypeNode $returnType, string $methodName, array $generics, array $parameters, string $description)
```
ondrejmirtes: This parameter needs to be added as last and needs to be optional, otherwise it's a BC break.

ondrejmirtes (changes requested):
And please fix the build failures :)

tyabu12:
@ondrejmirtes 
Thank you for quick reviewing!
I fixed your review comments and build failures.

src/Parser/PhpDocParser.php
```diff
@@ -346,6 +346,14 @@ private function parseMethodTagValue(TokenIterator $tokens): Ast\PhpDoc\MethodTa
 			exit;
 		}
 
+		$templateTypes = [];
+		if ($tokens->tryConsumeTokenType(Lexer::TOKEN_OPEN_ANGLE_BRACKET)) {
+			do {
+				$templateTypes[] = $this->parseMethodTagValueTemplateType($tokens);
```
ondrejmirtes: Could we call the already existing `parseTemplateTagValue`?

ondrejmirtes (changes requested):

src/Parser/PhpDocParser.php
tyabu12: @ondrejmirtes 
I fixed it.
`parseTemplateTagValue` also parse `description` text, so it was necessary to add the flag parameter  `$parseDescription`.

ondrejmirtes:
This is perfect, thank you very much :) I hope you follow-up with an implementation in phpstan-src :)

ondrejmirtes (approved):

ondrejmirtes:
I see that this was added in phpdoc-parser over a year ago, but you never followed up with implementing this in phpstan-src. Can you please do? Thanks.

PR #161, opened 2022-11-04
Add support for strict arrays and list shapes

danog:


ondrejmirtes:
Looks like Psalm didn't go with this syntax after all.

PR #165, opened 2022-11-15
In param tags variable name is optional

### LINKED ISSUE #164: [PSR-5] param tag with optional values
The param requires a paramname, however, this could be omitted when the param tag is describing just a single argument method or all other params are adequately documented. For example:

```
/**
 * @param int
 * /
 public function foo($arg): void 
{}
```

```
/**
 * @param int $arg
 * @param int Optional description
 * /
 public function foo($arg, $second): void 
{}
```
I do see this as edge cased, but it makes makes more sense when you align this with other tags for example the `@var` tag. Which has the same behavior. 

```
class Foo {
    /** @var int */
   public $bar
}
```

```
class Foo {
    /** 
     * @var int
     * @var string $foo
     */
   public $bar $foo
}
```

I'm working on a pr to support this.

jaapio:
Accoring to the PRS-5 standard paramnames are optional. The parser does not force docblocks to be fully valid, so it's ok to have multiple param tags without variable name. However this should never happen in real world docblocks.

Fixes #164

### DIFF
--- a/doc/grammars/phpdoc-param.peg
+++ b/doc/grammars/phpdoc-param.peg
@@ -1,5 +1,5 @@
 PhpDocParam
-  = AnnotationName Type? IsReference? IsVariadic? ParameterName Description?
+  = AnnotationName Type? IsReference? IsVariadic? ParameterName? Description?
 
 AnnotationName
   = '@param'
--- a/src/Parser/PhpDocParser.php
+++ b/src/Parser/PhpDocParser.php
@@ -266,9 +266,13 @@ class PhpDocParser
 			$type = $this->typeParser->parse($tokens);
 		}
 
+		if ($type !== null) {
+			$tokens->mustBePrecededByHorizontalWhitespace();
+		}
+
 		$isReference = $tokens->tryConsumeTokenType(Lexer::TOKEN_REFERENCE);
 		$isVariadic = $tokens->tryConsumeTokenType(Lexer::TOKEN_VARIADIC);
-		$parameterName = $this->parseRequiredVariableName($tokens);
+		$parameterName = $this->parseOptionalVariableName($tokens);
 		$description = $this->parseOptionalDescription($tokens);
 
 		if ($type !== null) {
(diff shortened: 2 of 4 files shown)


jaapio:
@ondrejmirtes a friendly reminder for this pr, is there more work required to get this merged?

ondrejmirtes:
Hi, I disagree with this one, it doesn't make sense to add now imho.

ondrejmirtes:
I see `@param` as a `parameterName => type` map. To allow entries without a parameter name doesn't make sense to me. PHPStan enforces this for many years, there isn't going to be much code left that looks like this.

jaapio:
Thanks for your response, I will try to solve this in a different way. As phpDocumentor still requires the support for this. How much I would love to remove this. 

It would help to remove this from psr-5. A discussion we are both part of. I will close this pr.

ondrejmirtes:
Is there any activity going around PSR-5? Seems dead and outdated to me at this point. I don't think we really need it :)

jaapio:
It's still alive. The main reason I think it makes sense is because we (phpDocumentor, phpstan and psalm) are all working in the same area. Where you are doing an awesome job, thank you very much for that.

I think when we bundle our strengths to make a good standard the whole community will benefit from that. Making documentation and static analyse go hand in hand together. One of the reasons I'm now trying to integrate this parser into our code. So we can support all types supported by phpstan. This will also give me the opportunity to help you improve this project.

Phpstorm has always followed phpDocumentor, and generics support was delayed because it was not part of psr-5. You can find this in their issue tracker. I'm sure that when we standardize things we will also have them on board to support all the awesomeness you are working on. 

I don't want to drag you into a political discussion, that would be a waist of our time. But I'm convinced of the fact that standards help to improve the adoption of new technologies. 

I wanted to have a chat with you about this topic at symfonycon. But I didn't have time to meet you somewhere. 

Thanks a again for helping me in my daily work with phpstan. And I hope we can have a chat about the possibilities in type strict php at some point.

ondrejmirtes:
But our tools move faster than any standard could. If the standard is ever "published", it will be obsolete in a month.

Right now the tools ecosystem is doing totally fine without any standard, including PhpStorm which follows what Psalm and PHPStan do.

Thank you for understanding too :)

PR #166, opened 2022-11-17
Allow method tag return type to be by reference

### LINKED ISSUE #158: Method tag is missing return by reference
The MethodTagValue doesn't contain an property returnsReference. This is not a very common option to use but we found a few of them in the wild. 

Full blown example will look like this:
  
```
/** @method static Type &myMethod() description here */
```

I will provide a PR later when ready with other things.

jaapio:
Like normal methods the method tag might return by reference. This fix introduces the support of reference return types. However in a real world implementation it not very likely one would ever mix reference return types with none reference return types. If even possible using php __call.

fixes #158

### DIFF
--- a/src/Parser/PhpDocParser.php
+++ b/src/Parser/PhpDocParser.php
@@ -7,6 +7,7 @@ use PHPStan\PhpDocParser\Ast\Type\IdentifierTypeNode;
 use PHPStan\PhpDocParser\Lexer\Lexer;
 use PHPStan\ShouldNotHappenException;
 use function array_key_exists;
+use function array_slice;
 use function array_values;
 use function count;
 use function trim;
@@ -330,6 +331,7 @@ class PhpDocParser
 	{
 		$isStatic = $tokens->tryConsumeTokenValue('static');
 		$returnTypeOrMethodName = $this->typeParser->parse($tokens);
+		$returnsReference = false;
 
 		if ($tokens->isCurrentTokenType(Lexer::TOKEN_IDENTIFIER)) {
 			$returnType = $returnTypeOrMethodName;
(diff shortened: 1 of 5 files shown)


src/Parser/PhpDocParser.php
```diff
@@ -340,6 +342,16 @@ private function parseMethodTagValue(TokenIterator $tokens): Ast\PhpDoc\MethodTa
 			$returnType = $isStatic ? new Ast\Type\IdentifierTypeNode('static') : null;
 			$methodName = $returnTypeOrMethodName->name;
 			$isStatic = false;
+		} elseif (
+			$returnTypeOrMethodName instanceof Ast\Type\IntersectionTypeNode
```
ondrejmirtes, line 346: No, that's a horrible hack 😂

jaapio: I'm sorry for this 😊

jaapio: How to solve this in a better way? I think this would require a change in the type parser as it is now to greedy.

jaapio:
@ondrejmirtes before I invest more time to improve this PR, would you merge this feature or is it not a valid case for phpstan? It would force me to find another solution on my side, but that's somehow okish.

ondrejmirtes:
@jaapio Hi, I'm pretty sceptical about this. I don't think that PHPStan would use this information in any way (that a method returns by reference). How do you think it'd enhance the analysis?

Also, what about just reading whether `__call` returns by reference, instead of changing the `@method` syntax?

jaapio:
I think from an analysis point of view, you could still read `__call`. But as a consumer of the parser not being phpstan, it makes sense, as the class containing the `@method` does not have to contain the implementation of `__call`. 

In case of phpstan the return type being a `&` makes sense, because a method will have side effects in that case. So you could warn people about mixing `@method` with and without return type by reference. The only way to know that if they added it by intention to `@method`.

ondrejmirtes:
I’m sorry, that doesn’t sound that exciting, unless you have a bug that’d be prevented with this in mind 😊

PR #167, opened 2022-11-18
Add specialized tags support

### LINKED ISSUE #157: [PSR-5] support colon in tag name
To make this parser better compatible with PSR-5 I would like to add support for a colon (`:`) in the tag name. As described: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#532-tag-specialization 

These specializations a not used in the wild, but it should be a small change allowing more projects to adopt this parser. 

Please let me know if you are open of a PR to add this.

jaapio:
Specialized tags are tags including a signle colon.

Fixes https://github.com/phpstan/phpdoc-parser/issues/157

### DIFF
--- a/src/Lexer/Lexer.php
+++ b/src/Lexer/Lexer.php
@@ -146,7 +146,7 @@ class Lexer
 
 			self::TOKEN_OPEN_PHPDOC => '/\\*\\*(?=\\s)\\x20?+',
 			self::TOKEN_CLOSE_PHPDOC => '\\*/',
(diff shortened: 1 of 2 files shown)


staabm:
what is the goal of this PR? why do you need support for such a prefix?

rvanvelzen:
@staabm This fixes #157

jaapio:
Sorry I forgot to add the ref to the issue :-), it makes this parser more compatible with psr-5. And usable for other tools following the draft standard.

ondrejmirtes:
Thank you. Gonna tag a release right away :)

PR #169, opened 2022-11-30
Support unsealed array shapes

jrmajor:
This PR adds support for `array{foo: int, ...}` syntax (phpstan/phpstan#8438).

### DIFF
--- a/src/Ast/Type/ArrayShapeNode.php
+++ b/src/Ast/Type/ArrayShapeNode.php
@@ -13,15 +13,25 @@ class ArrayShapeNode implements TypeNode
 	/** @var ArrayShapeItemNode[] */
 	public $items;
 
-	public function __construct(array $items)
+	/** @var bool */
+	public $sealed;
+
+	public function __construct(array $items, bool $sealed = true)
 	{
 		$this->items = $items;
+		$this->sealed = $sealed;
 	}
 
 
 	public function __toString(): string
 	{
-		return 'array{' . implode(', ', $this->items) . '}';
+		$items = $this->items;
+
+		if ($this->sealed) {
+			$items[] = '...';
+		}
+
+		return 'array{' . implode(', ', $items) . '}';
 	}
 
 }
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Thank you. This is useful so that unsealed array shapes in `@psalm-var` are not a parse error for PHPStan anymore, but I don't want to take any further action for this. To fully support all possible use-cases, PHPStan would need 3-4 different ConstantArrayType implementations, but right now we only have one.

src/Ast/Type/ArrayShapeNode.php
```diff
@@ -13,15 +13,25 @@ class ArrayShapeNode implements TypeNode
 	/** @var ArrayShapeItemNode[] */
 	public $items;
 
-	public function __construct(array $items)
+	/** @var bool */
+	public $sealed;
+
+	public function __construct(array $items, bool $sealed = true)
```
samsonasik, line 19: @jrmajor this cause error on Rector side, that's sealed true make all changes to add `...` which actually not, see

https://github.com/rectorphp/rector-src/actions/runs/3706328433/jobs/6281403238

we forced pin to use 1.5.0 and the issue disappear

/cc @TomasVotruba @ondrejmirtes

src/Parser/TypeParser.php
```diff
@@ -503,29 +503,32 @@ private function tryParseArrayOrOffsetAccess(TokenIterator $tokens, Ast\Type\Typ
...
 	{
 		$tokens->consumeTokenType(Lexer::TOKEN_OPEN_CURLY_BRACKET);
-		if ($tokens->tryConsumeTokenType(Lexer::TOKEN_CLOSE_CURLY_BRACKET)) {
-			return new Ast\Type\ArrayShapeNode([]);
-		}
 
-		$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
-		$items = [$this->parseArrayShapeItem($tokens)];
+		$items = [];
+		$sealed = true;
```
samsonasik, line 508: @jrmajor this cause error on Rector side, if this goes to `false` for default value, it may reduce the possible error.

that's sealed true make all changes to add `...` which actually not, see

https://github.com/rectorphp/rector-src/actions/runs/3706328433/jobs/6281403238

we forced pin to use 1.5.0 and the issue disappear

/cc @TomasVotruba @ondrejmirtes

src/Ast/Type/ArrayShapeNode.php
jrmajor: Sorry, that's actually a bug in `__toString()`. I'll provide a fix later today.

PR #171, opened 2022-12-15
Fix printing array shapes

jrmajor:
Fixes a bug reported in https://github.com/phpstan/phpdoc-parser/pull/169#discussion_r1050020908.

I've added tests for `__toString()` method on all nodes so that this doesn't happen again.
Along the way, I've also discovered and fixed minor whitespace issues with printing callable parameters.


ondrejmirtes:
Thank you!

samsonasik:
Thank you @jrmajor

PR #175, opened 2023-01-29
Add support for list shapes

zonuexe:
`strict-array` and `strict-list` were canceled, but list shapes was [introduced in Psalm 5](https://psalm.dev/articles/psalm-5).
It looks like you can merge some of the https://github.com/phpstan/phpdoc-parser/pull/161 patch by @danog. 

refs https://github.com/phpstan/phpstan/issues/8789

### DIFF
--- a/tests/PHPStan/Parser/TypeParserTest.php
+++ b/tests/PHPStan/Parser/TypeParserTest.php
@@ -669,6 +669,24 @@ class TypeParserTest extends TestCase
 				),
 			],
 			[
+				'list{
+				 	int,
+				 	string
+				 }',
+				new ArrayShapeNode([
+					new ArrayShapeItemNode(
+						null,
+						false,
+						new IdentifierTypeNode('int')
+					),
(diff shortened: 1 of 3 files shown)


tests/PHPStan/Parser/TypeParserTest.php
```diff
@@ -668,6 +668,24 @@ public function provideParseData(): array
 					Lexer::TOKEN_CLOSE_CURLY_BRACKET
 				),
 			],
+			[
+				'list{
+				 	int,
+				 	string
+				 }',
+				new ArrayShapeNode([
```
ondrejmirtes: Would be nice to know the array shape "kind" (array or list) at the ArrayShapeNode class.

ondrejmirtes (changes requested):

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #181, opened 2023-03-22
Update PHPCS

staabm:
as requested in https://github.com/phpstan/phpstan-strict-rules/pull/206#issuecomment-1475835384


ondrejmirtes:
Thank you!

PR #189, opened 2023-04-20
ConstExprParser: support numeric literal separator

jiripudil:
This PR adds support for an underscore separator in numeric literals, e.g.

```php
/** @var int<0, 999_999_999> */
```

The logic should be the same as supported by PHP (https://wiki.php.net/rfc/numeric_literal_separator). To maintain BC, this is also implemented similarly to how PHP does it: the separator is supported by the lexer but then stripped


doc/grammars/type.abnf
```diff
@@ -85,18 +85,18 @@ ConstantExpr
...
-	= ["-"] 1*ByteDecDigit "." *ByteDecDigit [ConstantFloatExp]
-	/ ["-"] 1*ByteDecDigit ConstantFloatExp
-	/ ["-"] "." 1*ByteDecDigit [ConstantFloatExp]
+	= ["-"] 1*ByteDecDigit *("_" 1*ByteDecDigit) "." [1*ByteDecDigit *("_" 1*ByteDecDigit)] [ConstantFloatExp]
+	/ ["-"] 1*ByteDecDigit *("_" 1*ByteDecDigit) ConstantFloatExp
+	/ ["-"] "." 1*ByteDecDigit *("_" 1*ByteDecDigit) [ConstantFloatExp]
 
 ConstantFloatExp
-	= "e" ["-"] 1*ByteDecDigit
+	= "e" ["-"] 1*ByteDecDigit *("_" 1*ByteDecDigit)
```
mvorisek, line 93: please see https://github.com/phpstan/phpstan-src/pull/2358#discussion_r1178091844, `+` before exponent should be supported, for consistency, also before any number.

ondrejmirtes:
Thank you very much! I'm gonna release this as part of phpdoc-parser 1.21 but it's not going to be right away, because I just did this massive amount of work https://github.com/phpstan/phpdoc-parser/commit/308c57c96db3237c6648ca266a72516d2be82f0a which I still need to improve a bit :)

jiripudil:
Sure, no rush :)

doc/grammars/type.abnf
jiripudil: yep, it makes sense to support the same range of expressions as PHP does. But it feels out of scope of this PR; feel free to follow up with a fix :)

mvorisek: I will, this PR is merged, thus there will be no conflicts... One question, why is the grammar defined on two places - doc/grammars/type.abnf and src/Lexer/Lexer.php, is one file generated from the other one?

ondrejmirtes: `doc/grammars/type.abnf` is mostly for documentation purposes and it's very likely out of sync. There's also FuzzyTest using it but I'm not sure about its purpose.

src/Parser/ConstExprParser.php
```diff
@@ -47,7 +48,7 @@ public function parse(TokenIterator $tokens, bool $trimStrings = false): Ast\Con
 
 			return $this->enrichWithAttributes(
 				$tokens,
-				new Ast\ConstExpr\ConstExprFloatNode($value),
+				new Ast\ConstExpr\ConstExprFloatNode(str_replace('_', '', $value)),
```
mvorisek, line 51: What is the reason for `str_replace` here? It seems there are no other text transformations done in general, for example `\d\.` is not normalized to `\d\.\d` etc.

jiripudil: Keeping the underscore here would be a BC break because phpstan-src simply casts this value to int or float, and `(int) '1_000_000'` evaluates to `1`.

PR #199, opened 2023-06-24
Add callable template support

mvorisek:
needed for https://github.com/phpstan/phpstan/issues/8964

based on https://github.com/phpstan/phpstan/issues/8964#issuecomment-1445362737 analysis

### DIFF
--- a/doc/grammars/type.abnf
+++ b/doc/grammars/type.abnf
@@ -35,7 +35,13 @@ GenericTypeArgument
     / TokenWildcard
 
 Callable
-	= TokenParenthesesOpen [CallableParameters] TokenParenthesesClose TokenColon CallableReturnType
+	= [CallableTemplate] TokenParenthesesOpen [CallableParameters] TokenParenthesesClose TokenColon CallableReturnType
+
+CallableTemplate
+	= TokenAngleBracketOpen CallableTemplateArgument *(TokenComma CallableTemplateArgument) TokenAngleBracketClose
+
+CallableTemplateArgument
+    = TokenIdentifier [1*ByteHorizontalWs TokenOf Type]
 
 CallableParameters
 	= CallableParameter *(TokenComma CallableParameter)
@@ -192,6 +198,9 @@ TokenIs
 TokenNot
 	= %s"not" 1*ByteHorizontalWs
 
+TokenOf
+	= %s"of" 1*ByteHorizontalWs
+
 TokenContravariant
     = %s"contravariant" 1*ByteHorizontalWs
 
@@ -211,7 +220,7 @@ TokenIdentifier
 
 ByteHorizontalWs
 	= %x09 ; horizontal tab
-	/ %x20 ; space
+	/ " "
 
 ByteNumberSign
 	= "+"
@@ -238,11 +247,8 @@ ByteIdentifierFirst
 	/ %x80-FF
 
 ByteIdentifierSecond
-	= %x30-39 ; 0-9
-	/ %x41-5A ; A-Z
-	/ "_"
-	/ %x61-7A ; a-z
-	/ %x80-FF
+	= ByteIdentifierFirst
+	/ %x30-39 ; 0-9
 
 ByteSingleQuote
 	= %x27    ; '
--- a/src/Ast/Type/CallableTypeNode.php
+++ b/src/Ast/Type/CallableTypeNode.php
@@ -13,6 +13,9 @@ class CallableTypeNode implements TypeNode
 	/** @var IdentifierTypeNode */
 	public $identifier;
 
+	/** @var CallableTypeTemplateNode[] */
+	public $templates;
+
 	/** @var CallableTypeParameterNode[] */
 	public $parameters;
 
@@ -21,12 +24,14 @@ class CallableTypeNode implements TypeNode
 
 	/**
 	 * @param CallableTypeParameterNode[] $parameters
+	 * @param CallableTypeTemplateNode[]  $templates
 	 */
-	public function __construct(IdentifierTypeNode $identifier, array $parameters, TypeNode $returnType)
+	public function __construct(IdentifierTypeNode $identifier, array $parameters, TypeNode $returnType, array $templates = [])
 	{
 		$this->identifier = $identifier;
 		$this->parameters = $parameters;
 		$this->returnType = $returnType;
+		$this->templates = $templates;
 	}
 
 
@@ -36,8 +41,11 @@ class CallableTypeNode implements TypeNode
 		if ($returnType instanceof self) {
 			$returnType = "({$returnType})";
 		}
+		$template = $this->templates !== []
+			? '<' . implode(', ', $this->templates) . '>'
+			: '';
 		$parameters = implode(', ', $this->parameters);
-		return "{$this->identifier}({$parameters}): {$returnType}";
+		return "{$this->identifier}{$template}({$parameters}): {$returnType}";
 	}
 
 }
--- a/src/Ast/Type/CallableTypeTemplateNode.php
+++ b/src/Ast/Type/CallableTypeTemplateNode.php
@@ -1,0 +1,35 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast\Type;
+
+use PHPStan\PhpDocParser\Ast\Node;
+use PHPStan\PhpDocParser\Ast\NodeAttributes;
+
+class CallableTypeTemplateNode implements Node
+{
+
+	use NodeAttributes;
+
+	/** @var IdentifierTypeNode */
+	public $identifier;
+
+	/** @var TypeNode|null */
+	public $bound;
+
+	public function __construct(IdentifierTypeNode $identifier, ?TypeNode $bound)
+	{
+		$this->identifier = $identifier;
+		$this->bound = $bound;
+	}
+
+	public function __toString(): string
+	{
+		$res = (string) $this->identifier;
+		if ($this->bound !== null) {
+			$res .= ' of ' . $this->bound;
+		}
+
+		return $res;
+	}
+
+}
--- a/src/Parser/TypeParser.php
+++ b/src/Parser/TypeParser.php
@@ -162,13 +162,17 @@ class TypeParser
 						return $type;
 					}
 
-					$type = $this->parseGeneric($tokens, $type);
+					$origType = $type;
+					$type = $this->tryParseCallable($tokens, $type, true);
+					if ($type === $origType) {
+						$type = $this->parseGeneric($tokens, $type);
 
-					if ($tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_SQUARE_BRACKET)) {
-						$type = $this->tryParseArrayOrOffsetAccess($tokens, $type);
+						if ($tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_SQUARE_BRACKET)) {
+							$type = $this->tryParseArrayOrOffsetAccess($tokens, $type);
+						}
 					}
 				} elseif ($tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_PARENTHESES)) {
-					$type = $this->tryParseCallable($tokens, $type);
+					$type = $this->tryParseCallable($tokens, $type, false);
 
 				} elseif ($tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_SQUARE_BRACKET)) {
 					$type = $this->tryParseArrayOrOffsetAccess($tokens, $type);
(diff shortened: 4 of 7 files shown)


src/Ast/Type/CallableTypeNode.php
```diff
@@ -22,7 +22,7 @@ class CallableTypeNode implements TypeNode
 	/**
 	 * @param CallableTypeParameterNode[] $parameters
 	 */
-	public function __construct(IdentifierTypeNode $identifier, array $parameters, TypeNode $returnType)
+	public function __construct(IdentifierTypeNode|GenericTypeNode $identifier, array $parameters, TypeNode $returnType)
```
mvorisek, line 25: @ondrejmirtes is this the right OOP design or you want separate `GenericCallableTypeNode`?

`GenericTypeNode` is a litttle problematic then, as `$variances` are stored separately from `$genericTypes` there, shouldn't that be stored in another separate class?

mvorisek, line 25: @ondrejmirtes is this the right OOP design or you want separate `GenericCallableTypeNode`?

`GenericTypeNode` is a litttle problematic then, as `$variances` are stored separately from `$genericTypes` there, shouldn't that be stored in another separate class? /cc @jiripudil

doc/grammars/type.abnf
```diff
@@ -35,7 +35,13 @@ GenericTypeArgument
     / TokenWildcard
 
 Callable
-	= TokenParenthesesOpen [CallableParameters] TokenParenthesesClose TokenColon CallableReturnType
+	= [CallableTemplate] TokenParenthesesOpen [CallableParameters] TokenParenthesesClose TokenColon CallableReturnType
```
mvorisek, line 38: @ondrejmirtes can I ask you for a hint how to fix the fuzzer tests? The grammar seems correct to me, and the manualy tests are passing as expected...

src/Parser/TypeParser.php
```diff
@@ -476,7 +484,65 @@ private function parseCallable(TokenIterator $tokens, Ast\Type\IdentifierTypeNod
...
+		$identifier = $this->enrichWithAttributes(
+			$tokens,
+			new Ast\Type\IdentifierTypeNode($tokens->currentTokenValue()),
+			$startLine,
+			$startIndex
+		);
+		$tokens->consumeTokenType(Lexer::TOKEN_IDENTIFIER);
+
+		$ofType = null;
+		if ($tokens->tryConsumeTokenValue('of')) {
```
mvorisek, line 544: like in https://github.com/phpstan/phpdoc-parser/blob/1.22.0/src/Parser/PhpDocParser.php#L908

mvorisek:
PR is done.

src/Ast/Type/CallableTypeTemplateNode.php
```diff
@@ -0,0 +1,34 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast\Type;
+
+use PHPStan\PhpDocParser\Ast\NodeAttributes;
+
+class CallableTypeTemplateNode implements TypeNode
```
jiripudil: nitpick: I don't think this should be a `TypeNode`, just `Node`

src/Parser/TypeParser.php
```diff
@@ -381,42 +385,33 @@ public function isHtml(TokenIterator $tokens): bool
...
+		$startIndex = $baseType->getAttribute(Ast\Attribute::START_INDEX);
 		$genericTypes = [];
 		$variances = [];
 
-		[$genericTypes[], $variances[]] = $this->parseGenericTypeArgument($tokens);
-
-		$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
-
-		while ($tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA)) {
+		$isFirst = true;
```
jiripudil: Is there a benefit to this change, such as in performance? Personally, I find it more difficult to read this way. Feels like I have to think harder to figure out what's going on here. I understand that's very subjective, just wanted to raise the question :)

src/Parser/TypeParser.php
```diff
@@ -476,7 +475,65 @@ private function parseCallable(TokenIterator $tokens, Ast\Type\IdentifierTypeNod
...
+		return new Ast\Type\CallableTypeNode($identifier, $parameters, $returnType, $templates);
+	}
+
+
+	/**
+	 * @return Ast\Type\CallableTypeTemplateNode[]
+	 *
+	 * @phpstan-impure
+	 */
+	private function parseCallableTemplates(TokenIterator $tokens): array
```
jiripudil, line 504: I'm wondering if this should be tied to `CallableType`? I believe the concept is more universal, and could later be reused in other places where an inline declaration of template types might be useful, such as in generic type aliases 🤔

jiripudil (commented):
Hi, I took a look since this somehow ended up in my notifications and piqued my curiosity. But I'm really just passing by and only wanted to point out a few topics for discussion, so don't take my comments too seriously :)

Nice work overall!

src/Parser/TypeParser.php
mvorisek: Previously some code parts were duplicated...

Now the LOC is lower and IMO it is more readable too. In first iteration, we parse/expect no comma, in 2nd+ we parse it with delimiting comma + allow trailing comma.

src/Ast/Type/CallableTypeTemplateNode.php
mvorisek: Probably yes, as the class cannot be used as type itself. And it can be always, with full BC, changed back anytime ;-)

src/Parser/TypeParser.php
mvorisek: So you prefer to call it "template" only? @ondrejmirtes should decide, as it might be confusing with "generic" later...

Can you post any phpdoc example, outside callable, where a template can make sense?

jiripudil: Haven't thought about the naming, honestly. As for the example, as I've said, generic type aliases could be a candidate:

```php
/**
 * @phpstan-type Factory<T> = callable(): T
 */
```

I'm not planning to implement them anytime soon (if at all), but it shows that the concept of inline template type declarations could be reused. That being said, it can easily be refactored when/if the need arises

mvorisek:
@ondrejmirtes this parser PR should be fully done, https://github.com/phpstan/phpstan/issues/8964#issuecomment-1445362737 might be easy for you, but I am not familiar with phpstan internals so I would appreciate a help with the integration into phpstan. Thank you.

mvorisek:
@ondrejmirtes this PR was here for several weeks, may I ask you to review it?

mvorisek:
@ondrejmirtes this PR is here for over a month, may I ask you to review it?

src/Printer/Printer.php
```diff
@@ -370,10 +371,13 @@ private function printType(TypeNode $node): string
 			} else {
 				$returnType = $this->printType($node->returnType);
 			}
+			$template = $node->templates !== []
+				? '<' . implode(', ', $node->templates) . '>'
```
ondrejmirtes: Types need to be printed using `$this->printType` here, however here you're printing them using `TypeNode::__toString()`.

src/Printer/Printer.php
```diff
@@ -97,6 +97,7 @@ final class Printer
 		ArrayShapeNode::class . '->items' => ', ',
 		ObjectShapeNode::class . '->items' => ', ',
 		CallableTypeNode::class . '->parameters' => ', ',
+		CallableTypeNode::class . '->templates' => ', ',
```
ondrejmirtes, line 101: 1) This change is not tested.
2) I believe we need to test other situations. Like when gets `CallableTypeNode` added or removed, we also need to add `<...>` and remove `<...>` around them.

ondrejmirtes (changes requested):
These are the two main issues I see in this PR.

Also I find the changes of the parser hard to understand, I have to inspect them in depth to understand the implications. Maybe if you could do the changes to the TypeParser in multiple smaller steps, I'd welcome them as separate PRs that I would merge before this one.

The credo to live by is: "Make the change easy, then make the easy change."

src/Printer/Printer.php
mvorisek: what test should I use as a base to start?

ondrejmirtes: All the tests in PrinterTest, for example:

```php
		yield [
			'/** @var string&(integer|float) */',
			'/** @var string&(int|float) */',
			new class extends AbstractNodeVisitor {

				public function enterNode(Node $node)
				{
					if ($node instanceof IdentifierTypeNode && $node->name === 'integer') {
						$node->name = 'int';
					}

					return $node;
				}

			},
		];
```

This test accepts an original PHPDoc, modifies it using the visitor, and then asserts that the printer correctly prints the modified PHPDoc.

src/Printer/Printer.php
mvorisek: done

src/Printer/Printer.php
mvorisek: > This change is not tested.

https://github.com/phpstan/phpdoc-parser/blob/1.24.2/src/Printer/Printer.php#L100 is not tested either, please add tests or please loosen this requirement for this contribution.

src/Ast/Type/CallableTypeTemplateNode.php
```diff
@@ -0,0 +1,35 @@
...
+
+use PHPStan\PhpDocParser\Ast\Node;
+use PHPStan\PhpDocParser\Ast\NodeAttributes;
+
+class CallableTypeTemplateNode implements Node
+{
+
+	use NodeAttributes;
+
+	/** @var IdentifierTypeNode */
```
mvorisek, line 13: `TemplateTagValueNode` names it `name` of type `string`, @ondrejmirtes do you want apply any of these?

src/Printer/Printer.php
mvorisek: > I believe we need to test other situations. Like when gets CallableTypeNode added or removed, we also need to add <...> and remove <...> around them.

addressed

mvorisek:
All feedback should be addressed.

mvorisek:
@ondrejmirtes the feedback should be addressed, if there is anything else please let me know.

mvorisek:
@ondrejmirtes sorry to tagging you again, can I please ask you to merge this PR if there is no feedback left?

src/Parser/TypeParser.php
```diff
@@ -484,7 +492,65 @@ private function parseCallable(TokenIterator $tokens, Ast\Type\IdentifierTypeNod
...
+		$tokens->consumeTokenType(Lexer::TOKEN_OPEN_ANGLE_BRACKET);
+
+		$templates = [];
+
+		$isFirst = true;
+		while ($isFirst || $tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA)) {
+			$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
+
+			// trailing comma case
+			if (!$isFirst && $tokens->isCurrentTokenType(Lexer::TOKEN_CLOSE_ANGLE_BRACKET)) {
```
ondrejmirtes, line 515: I don't understand why is there so much repeated code. We're already parsing template types in similar position for `@method`, there has to be a way to reuse that code: https://github.com/phpstan/phpdoc-parser/pull/160

Everything you should need is already in parseTemplateTagValue.

ondrejmirtes (changes requested):

src/Parser/TypeParser.php
mvorisek: Can you please guide me how and what to reuse? The code in #160 (Parser/PhpDocParser in general) is for phpdoc instead of for type (Parser/TypeParser).

If the newly introduced Ast/Type/CallableTypeTemplateNode class is wanted (and it probably is - callable template should not consist of anything like description), then the code cannot be simplified much. As the grammar is not completely the same, basically the only parsing part to deduplicate would be `<`, `,`, `of`.

If you want, feel free to push any changes to my branch.

jiripudil: I think #160 implements the very same use case as this PR: it parses inline template tag definitions, disallowing descriptions, but accounting for everything else (bound types, defaults, ...). Perhaps it could suffice to extract the `parseTemplateTagValue` method into a `TemplateTagParser` that could be used in `TypeParser` as well?

ondrejmirtes: Yes, that logic should be extracted so that it's usable by both TypeParser and PhpDocParser.

mad-briller: @mvorisek are you going to look into this feedback or should i have a go at it?

ondrejmirtes: @mad-briller Please have a go at it :)

mvorisek: @mad-briller would you like to join the forces? Feel free to either reuse this PR as a separate one or submit a PR againts https://github.com/mvorisek/phpdoc-parser/tree/parse_generic_callable - this PR should be basically done, but the code should be deduplicated per https://github.com/phpstan/phpdoc-parser/pull/199#discussion_r1375390544 request.

mvorisek:
Landed thru #232, thanks everyone involved!

PR #202, opened 2023-07-15
Support "one-line" comments (starting with //)

### LINKED ISSUE #184: Multiline array shapes with 'comments'
# Feature request

For types that involve more complex associative arrays, it is nice to know what each of key/value pairs are for. Does there exist a way to have 'type comments' that allow you to insert comments/descriptions in multiline array types?

For example, below is an associative array that has a bunch of expected key values, along with 'comments' to help describe what each key is for. Is there a standard for psalm to do this.

```php
/**
 *
 * @var array<
 *  string,                   // name of the rule
 *  array{
 *   params: array<           // list of parameters defined by the rule
 *    array{
 *     key: string,           // the key used to identify the value in the POST
 *     nullOK: bool,          // the key used to identify the value in the POST
 *     blankOK: bool,         // whether the param's value can be an empty string
 *     type: string,          // the type of value
 *     regExp?: string,       // regular expression used for validation if the type is 'regex'
 *     context?: string,      // extra data required for validation (currenlty only used if type is 'signed')
 *     name?: string,         // human readable name for the param (used by error messages, defaults to 'key')
 *     description?: string,  // a description of the parameter
 *     list?: bool,           // whether the param's value is a list
 *     delim?: string,        // delimeter to split the list on (e.g. [val => 'this,is,a,list', delim => ','])
 *     trim?: string,         // whether to strip whitespace from elements in list
 *     trimLevel?: string,    // 0 to trim whitespace from beginning and end of items,
 *     minErr?: int,          // minimum length of the param's value
 *     maxErr?: int,          // maximum length of the param's value
 *     allowedVals?: array,   // a list of allowed values
 *     blackList?: array      // a list of blacklisted/blocked values
 *    }>,
 *   handlers: array<
 *    string,                 // name of the handler
 *    array{
 *     func: string,          // the name of the function in the app's validator (e.g. '\ESP\app\Validator::func')
 *     extraArgs?: array,     // any parameters that are not $action, $post, or $postIn - ORDER MATTERS
 *     return?: bool          // whether error's are returned as a json or an exception is thrown (defaults to false)
 *    }>
 * }>
 */
public static $rules = [];
```
This would be incredibly useful. Thanks!

rhertogh:
This PR adds support for comments inside type definitions, e.g.:

```php
/** @return array { // Array with comments.
 *     // Comments can be placed on their own line. 
 *     foo: string, // An array key description.
 *     bar: array {, // Another array key description.
 *         'foo//bar': string, // Array key with "//" in it's name.
 *     },
 * }
 */
```

Fixes #184

### DIFF
--- a/src/Parser/TokenIterator.php
+++ b/src/Parser/TokenIterator.php
@@ -9,6 +9,7 @@ use function assert;
 use function count;
 use function in_array;
 use function strlen;
+use function strpos;
 use function substr;
 
 class TokenIterator
--- a/src/Parser/TokenIterator.php
+++ b/src/Parser/TokenIterator.php
@@ -380,4 +380,37 @@ class TokenIterator
 			&& $this->hasTokenImmediatelyAfter($endPos, Lexer::TOKEN_CLOSE_PARENTHESES);
 	}
 
+	/**
+	 * Strip PHP style "one-line" comments (text starting with //)
+	 */
+	public function stripComments(): void
+	{
+		$line = 1;
+		$cleanTokens = [];
+		for ($i = 0; $i < count($this->tokens); $i++) {
+			$token = $this->tokens[$i];
+			if (
+				$token[Lexer::TYPE_OFFSET] === Lexer::TOKEN_OTHER
+				&& strpos($token[Lexer::VALUE_OFFSET], '//') === 0
+			) {
+				while (
+					strpos($this->tokens[$i][Lexer::VALUE_OFFSET], "\n") === false
+					&& $i < count($this->tokens)
+				) {
+					$i++;
+				}
+				if ($this->tokens[$i][Lexer::LINE_OFFSET] !== $line) {
+					for ($j = $i + 1; $j < count($this->tokens); $j++) {
+						$this->tokens[$j][Lexer::LINE_OFFSET]--;
+					}
+				}
+			} else {
+				$cleanTokens[] = $token;
+				$line = $token[Lexer::LINE_OFFSET];
+			}
+		}
+
+		$this->tokens = $cleanTokens;
+	}
+
 }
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Hi, thank you, I have a few points that will makes this more complicated.

1) The comment needs to be available for each type node in the node attributes.
2) The comments need to be correctly preserved for the format-preserving printer, meaning it will get a bit more complicated. Fortunately it's heavily inspired by the printer from nikic/PHP-Parser so we have some starting point.

rhertogh:
@ondrejmirtes 
> 1. The comment needs to be available for each type node in the node attributes.
> 2. The comments need to be correctly preserved for the format-preserving printer, meaning it will get a bit more complicated. Fortunately it's heavily inspired by the printer from nikic/PHP-Parser so we have some starting point.

Could you point me to the locations where this should be implemented, then I'll give it a shot.

ondrejmirtes:
Comments need to be added as attributes similarly to how `Attribute::START_LINE` is added today. See calls to methods called `enrichWithAttributes`.

As for Printer, see tests in PrinterTest. Comments should be preserved correctly when printing the same PHPDoc again, and correct comment should be removed when for example removing a key from an array shape.

shmax:
@rhertogh do you have any plans to proceed with this PR? I think it's an important feature... 🙏

shmax:
@ondrejmirtes  

In the event that @rhertogh is MIA, any chance I can prevail upon you to implement this? If you're too busy I can take a crack at it, but I have to confess I couldn't make much sense of the printer tests...

rhertogh:
@shmax Haven't found the time to continue working on this😅. So if you want to take a swing at it please feel free to do so.

shmax:
Okay, will do. 👍

ondrejmirtes:
Superseded by https://github.com/phpstan/phpdoc-parser/pull/264

PR #206, opened 2023-07-27
[PhpDoc] Add SpacelessPhpDocTagNode

samsonasik:
@ondrejmirtes since the `Printer` class is `final`, in rector-src for usecase of our custom `SpacelessPhpDocTagNode` for FQCN doctrine use, I add this to the `PhpDoc` namespace so can be used on rector-src to avoid space:

```diff
- * @Table ("Table_Name")
+ * @Table("Table_Name")
```

https://github.com/rectorphp/rector-src/pull/4573#discussion_r1274786024 . We can't use `parseDoctrineAnnotations` flag on extended `PhpDocParser` at 

https://github.com/rectorphp/rector-src/blob/729f03b6e9abccb7c4d7a4515e4dfed61e725e71/packages/BetterPhpDocParser/PhpDocParser/BetterPhpDocParser.php#L57-L58

as it cause invalid output:

```diff
 /**
- * @Table("Table_Name")
+ * @Doctrine\ORM\Mapping\Table("Table_Name")
```

### DIFF
--- a/src/Ast/PhpDoc/SpacelessPhpDocTagNode.php
+++ b/src/Ast/PhpDoc/SpacelessPhpDocTagNode.php
@@ -1,0 +1,13 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast\PhpDoc;
+
+use Stringable;
+
+final class SpacelessPhpDocTagNode extends PhpDocTagNode implements Stringable
+{
+    public function __toString(): string
+    {
+        return $this->name . $this->value;
+    }
+}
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Hi, this does not make sense for phpdoc-parser. It already supports Doctrine annotations parsing and printing.

Rector should adopt the way phpdoc-parser parses Doctrine annotations, meaning all of this code could be deleted (including SpacelessPhpDocTagNode):

* https://github.com/rectorphp/rector-src/blob/main/packages/BetterPhpDocParser/PhpDocParser/DoctrineAnnotationDecorator.php
* https://github.com/rectorphp/rector-src/blob/main/packages/BetterPhpDocParser/PhpDocParser/StaticDoctrineAnnotationParser.php
* https://github.com/rectorphp/rector-src/tree/main/packages/BetterPhpDocParser/PhpDoc

The reason why Rector currently uses SpacelessPhpDocTagNode is instead solved here: https://github.com/phpstan/phpdoc-parser/blob/4a1ab8e11e9957f9cc9f89f87a7c912489f08119/src/Ast/PhpDoc/PhpDocTagNode.php#L29-L31

As an alternative solution, you can use your current SpacelessPhpDocTagNode and transform it into something phpdoc-parser understands, use these classes: https://github.com/phpstan/phpdoc-parser/tree/1.23.x/src/Ast/PhpDoc/Doctrine

samsonasik:
@ondrejmirtes ok, I will try, thank you for the suggestion 👍

PR #209, opened 2023-09-07
Allow asserting the type of `$this`

ekisu:
Closes phpstan/phpstan#8904

I've also added some tests on the PHPStan side, to ensure the type assertions work properly: phpstan/phpstan-src#2609


ondrejmirtes:
Perfect, thank you!

PR #213, opened 2023-09-17
support comments in array shape

shmax:
This is my first experiment in trying to support comments in array shape:

```php
array{
    // a is for apple
	a: int,
}

array{
    // a is for apple
    // a is also for aardvark
	a: int,
}

```

Note: as pointed out by @ondrejmirtes , multi-line comments within a block comment aren't valid PHP, so single line comments like this are the best we can do.

### DIFF
--- a/src/Ast/Attribute.php
+++ b/src/Ast/Attribute.php
@@ -13,4 +13,6 @@ final class Attribute
 
 	public const ORIGINAL_NODE = 'originalNode';
 
+	public const COMMENTS = 'comments';
+
 }
--- a/src/Ast/Comment.php
+++ b/src/Ast/Comment.php
@@ -1,0 +1,31 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast;
+
+use function trim;
+
+class Comment
+{
+
+	/** @var string */
+	public $text;
+
+	/** @var int */
+	public $startLine;
+
+	/** @var int */
+	public $startIndex;
+
+	public function __construct(string $text, int $startLine = -1, int $startIndex = -1)
+	{
+		$this->text = $text;
+		$this->startLine = $startLine;
+		$this->startIndex = $startIndex;
+	}
+
+	public function getReformattedText(): ?string
+	{
+		return trim($this->text);
+	}
+
+}
--- a/src/Lexer/Lexer.php
+++ b/src/Lexer/Lexer.php
@@ -50,6 +50,8 @@ class Lexer
 	public const TOKEN_NEGATED = 35;
 	public const TOKEN_ARROW = 36;
 
+	public const TOKEN_COMMENT = 37;
+
 	public const TOKEN_LABELS = [
 		self::TOKEN_REFERENCE => '\'&\'',
 		self::TOKEN_UNION => '\'|\'',
@@ -65,6 +67,7 @@ class Lexer
 		self::TOKEN_OPEN_CURLY_BRACKET => '\'{\'',
 		self::TOKEN_CLOSE_CURLY_BRACKET => '\'}\'',
 		self::TOKEN_COMMA => '\',\'',
+		self::TOKEN_COMMENT => '\'//\'',
 		self::TOKEN_COLON => '\':\'',
 		self::TOKEN_VARIADIC => '\'...\'',
 		self::TOKEN_DOUBLE_COLON => '\'::\'',
@@ -160,6 +163,7 @@ class Lexer
 			self::TOKEN_CLOSE_CURLY_BRACKET => '\\}',
 
 			self::TOKEN_COMMA => ',',
+			self::TOKEN_COMMENT => '((?<![:/])\/\/[^\n]*)',
 			self::TOKEN_VARIADIC => '\\.\\.\\.',
 			self::TOKEN_DOUBLE_COLON => '::',
 			self::TOKEN_DOUBLE_ARROW => '=>',
--- a/src/Parser/TokenIterator.php
+++ b/src/Parser/TokenIterator.php
@@ -3,6 +3,7 @@
 namespace PHPStan\PhpDocParser\Parser;
 
 use LogicException;
+use PHPStan\PhpDocParser\Ast\Comment;
 use PHPStan\PhpDocParser\Lexer\Lexer;
 use function array_pop;
 use function assert;
@@ -17,6 +18,9 @@ class TokenIterator
 	/** @var list<array{string, int, int}> */
 	private $tokens;
 
+	/** @var array<Comment> */
+	private $comments = [];
+
 	/** @var int */
 	private $index;
 
@@ -24,7 +28,9 @@ class TokenIterator
 	private $savePoints = [];
 
 	/** @var list<int> */
-	private $skippedTokenTypes = [Lexer::TOKEN_HORIZONTAL_WS];
+	private $skippedTokenTypes = [
+		Lexer::TOKEN_HORIZONTAL_WS,
+		Lexer::TOKEN_COMMENT];
 
 	/** @var string|null */
 	private $newline = null;
@@ -154,8 +160,7 @@ class TokenIterator
 			}
 		}
 
-		$this->index++;
-		$this->skipIrrelevantTokens();
+		$this->next();
 	}
 
 
@@ -168,8 +173,7 @@ class TokenIterator
 			$this->throwError($tokenType, $tokenValue);
 		}
 
-		$this->index++;
-		$this->skipIrrelevantTokens();
+		$this->next();
 	}
 
 
@@ -180,12 +184,30 @@ class TokenIterator
 			return false;
 		}
 
-		$this->index++;
-		$this->skipIrrelevantTokens();
+		$this->next();
 
 		return true;
 	}
 
+	/**
+	 * @return Comment[]
+	 */
+	public function flushComments(): array
+	{
+		$res = $this->comments;
+		$this->comments = [];
+		return $res;
+	}
+
+	/** @phpstan-impure */
+	public function tryConsumeTokenTypeAll(int $tokenType): bool
+	{
+		$found = false;
+		while ($this->tryConsumeTokenType($tokenType)) {
+			$found = true;
+		}
+		return $found;
+	}
 
 	/** @phpstan-impure */
 	public function tryConsumeTokenType(int $tokenType): bool
@@ -200,8 +222,7 @@ class TokenIterator
 			}
 		}
 
-		$this->index++;
-		$this->skipIrrelevantTokens();
+		$this->next();
 
 		return true;
 	}
@@ -256,6 +277,11 @@ class TokenIterator
 			if (!isset($this->tokens[$this->index + 1])) {
 				break;
 			}
+
+			if ($this->currentTokenType() === Lexer::TOKEN_COMMENT) {
+				$this->comments[] = new Comment($this->currentTokenValue(), $this->currentTokenLine(), $this->currentTokenIndex());
+			}
+
 			$this->index++;
 		}
 	}
@@ -298,7 +324,6 @@ class TokenIterator
 		assert($index !== null);
 		$this->index = $index;
 	}
-
 
 	/**
 	 * @throws ParserException
--- a/src/Parser/TypeParser.php
+++ b/src/Parser/TypeParser.php
@@ -23,10 +23,17 @@ class TypeParser
 	private $useLinesAttributes;
 
 	/** @var bool */
+	private $useCommentsAttributes;
+
+	/** @var bool */
 	private $useIndexAttributes;
 
 	/**
-	 * @param array{lines?: bool, indexes?: bool} $usedAttributes
+	 * @param array{
+	 *     lines?: bool,
+	 *     indexes?: bool,
+	 *     comments?: bool
+	 * } $usedAttributes
 	 */
 	public function __construct(
 		?ConstExprParser $constExprParser = null,
@@ -38,6 +45,7 @@ class TypeParser
 		$this->quoteAwareConstExprString = $quoteAwareConstExprString;
 		$this->useLinesAttributes = $usedAttributes['lines'] ?? false;
 		$this->useIndexAttributes = $usedAttributes['indexes'] ?? false;
+		$this->useCommentsAttributes = $usedAttributes['comments'] ?? false;
 	}
 
 	/** @phpstan-impure */
@@ -66,6 +74,7 @@ class TypeParser
 	 * @internal
 	 * @template T of Ast\Node
 	 * @param T $type
+	 *
 	 * @return T
 	 */
 	public function enrichWithAttributes(TokenIterator $tokens, Ast\Node $type, int $startLine, int $startIndex): Ast\Node
@@ -73,6 +82,10 @@ class TypeParser
 		if ($this->useLinesAttributes) {
 			$type->setAttribute(Ast\Attribute::START_LINE, $startLine);
 			$type->setAttribute(Ast\Attribute::END_LINE, $tokens->currentTokenLine());
+		}
+
+		if ($this->useCommentsAttributes) {
+			$type->setAttribute(Ast\Attribute::COMMENTS, $tokens->flushComments());
 		}
 
 		if ($this->useIndexAttributes) {
@@ -310,7 +323,7 @@ class TypeParser
 
 		$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
 		$tokens->consumeTokenType(Lexer::TOKEN_NULLABLE);
-		$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
+		$tokens->tryConsumeTokenTypeAll(Lexer::TOKEN_PHPDOC_EOL);
 
 		$ifType = $this->parse($tokens);
 
@@ -398,41 +411,33 @@ class TypeParser
 	public function parseGeneric(TokenIterator $tokens, Ast\Type\IdentifierTypeNode $baseType): Ast\Type\GenericTypeNode
 	{
 		$tokens->consumeTokenType(Lexer::TOKEN_OPEN_ANGLE_BRACKET);
-		$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
+		$tokens->tryConsumeTokenTypeAll(Lexer::TOKEN_PHPDOC_EOL);
 
+		$startLine = $baseType->getAttribute(Ast\Attribute::START_LINE);
+		$startIndex = $baseType->getAttribute(Ast\Attribute::START_INDEX);
 		$genericTypes = [];
 		$variances = [];
 
-		[$genericTypes[], $variances[]] = $this->parseGenericTypeArgument($tokens);
-
-		$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
-
-		while ($tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA)) {
+		$isFirst = true;
+		while ($isFirst || $tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA)) {
 			$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
-			if ($tokens->tryConsumeTokenType(Lexer::TOKEN_CLOSE_ANGLE_BRACKET)) {
-				// trailing comma case
-				$type = new Ast\Type\GenericTypeNode($baseType, $genericTypes, $variances);
-				$startLine = $baseType->getAttribute(Ast\Attribute::START_LINE);
-				$startIndex = $baseType->getAttribute(Ast\Attribute::START_INDEX);
-				if ($startLine !== null && $startIndex !== null) {
-					$type = $this->enrichWithAttributes($tokens, $type, $startLine, $startIndex);
-				}
 
-				return $type;
+			// trailing comma case
+			if (!$isFirst && $tokens->isCurrentTokenType(Lexer::TOKEN_CLOSE_ANGLE_BRACKET)) {
+				break;
 			}
+			$isFirst = false;
+
 			[$genericTypes[], $variances[]] = $this->parseGenericTypeArgument($tokens);
 			$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
 		}
 
-		$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
-		$tokens->consumeTokenType(Lexer::TOKEN_CLOSE_ANGLE_BRACKET);
-
 		$type = new Ast\Type\GenericTypeNode($baseType, $genericTypes, $variances);
-		$startLine = $baseType->getAttribute(Ast\Attribute::START_LINE);
-		$startIndex = $baseType->getAttribute(Ast\Attribute::START_INDEX);
 		if ($startLine !== null && $startIndex !== null) {
 			$type = $this->enrichWithAttributes($tokens, $type, $startLine, $startIndex);
 		}
+
+		$tokens->consumeTokenType(Lexer::TOKEN_CLOSE_ANGLE_BRACKET);
 
 		return $type;
 	}
@@ -477,7 +482,7 @@ class TypeParser
 			$parameters[] = $this->parseCallableParameter($tokens);
 			$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
 			while ($tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA)) {
-				$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
+				$tokens->tryConsumeTokenTypeAll(Lexer::TOKEN_PHPDOC_EOL);
 				if ($tokens->isCurrentTokenType(Lexer::TOKEN_CLOSE_PARENTHESES)) {
 					break;
 				}
@@ -533,7 +538,7 @@ class TypeParser
 			return $this->parseNullable($tokens);
 
 		} elseif ($tokens->tryConsumeTokenType(Lexer::TOKEN_OPEN_PARENTHESES)) {
-			$type = $this->parse($tokens);
+			$type = $this->subParse($tokens);
 			$tokens->consumeTokenType(Lexer::TOKEN_CLOSE_PARENTHESES);
 			if ($tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_SQUARE_BRACKET)) {
 				$type = $this->tryParseArrayOrOffsetAccess($tokens, $type);
@@ -748,6 +753,8 @@ class TypeParser
 		$items = [];
 		$sealed = true;
 
+		$done = false;
+
 		do {
 			$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
 
@@ -761,10 +768,19 @@ class TypeParser
 				break;
 			}
 
-			$items[] = $this->parseArrayShapeItem($tokens);
-
+			$item = $this->parseArrayShapeItem($tokens);
+			$items[] = $item;
 			$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
-		} while ($tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA));
+			if (!$tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA)) {
+				$done = true;
+			}
+			if ($tokens->currentTokenType() !== Lexer::TOKEN_COMMENT) {
+				continue;
+			}
+
+			$tokens->next();
+
+		} while (!$done);
 
 		$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
 		$tokens->consumeTokenType(Lexer::TOKEN_CLOSE_CURLY_BRACKET);
@@ -778,12 +794,17 @@ class TypeParser
 	{
 		$startLine = $tokens->currentTokenLine();
 		$startIndex = $tokens->currentTokenIndex();
+
+		// parse any comments above the item
+		$tokens->tryConsumeTokenTypeAll(Lexer::TOKEN_PHPDOC_EOL);
+
 		try {
 			$tokens->pushSavePoint();
 			$key = $this->parseArrayShapeKey($tokens);
 			$optional = $tokens->tryConsumeTokenType(Lexer::TOKEN_NULLABLE);
 			$tokens->consumeTokenType(Lexer::TOKEN_COLON);
 			$value = $this->parse($tokens);
+
 			$tokens->dropSavePoint();
 
 			return $this->enrichWithAttributes(
@@ -881,12 +902,19 @@ class TypeParser
 		$startLine = $tokens->currentTokenLine();
 		$startIndex = $tokens->currentTokenIndex();
 
+		$tokens->tryConsumeTokenTypeAll(Lexer::TOKEN_PHPDOC_EOL);
+
 		$key = $this->parseObjectShapeKey($tokens);
 		$optional = $tokens->tryConsumeTokenType(Lexer::TOKEN_NULLABLE);
 		$tokens->consumeTokenType(Lexer::TOKEN_COLON);
 		$value = $this->parse($tokens);
 
-		return $this->enrichWithAttributes($tokens, new Ast\Type\ObjectShapeItemNode($key, $optional, $value), $startLine, $startIndex);
+		return $this->enrichWithAttributes(
+			$tokens,
+			new Ast\Type\ObjectShapeItemNode($key, $optional, $value),
+			$startLine,
+			$startIndex
+		);
 	}
 
 	/**
--- a/tests/PHPStan/Parser/PhpDocParserTest.php
+++ b/tests/PHPStan/Parser/PhpDocParserTest.php
@@ -4563,21 +4563,56 @@ some text in the middle'
 		];
 
 		yield [
-			'invalid $this',
+			'OK $this',
 			'/** @phpstan-assert Type $this */',
 			new PhpDocNode([
 				new PhpDocTagNode(
 					'@phpstan-assert',
-					new InvalidTagValueNode(
-						'Type $this',
-						new ParserException(
-							'*/',
-							Lexer::TOKEN_CLOSE_PHPDOC,
-							31,
-							Lexer::TOKEN_ARROW,
-							null,
-							1
-						)
+					new AssertTagValueNode(
+						new IdentifierTypeNode('Type'),
+						'$this',
+						false,
+						''
+					)
+				),
+			]),
+		];
+
+		yield [
+			'OK $this with description',
+			'/** @phpstan-assert Type $this assert Type to $this */',
+			new PhpDocNode([
+				new PhpDocTagNode(
+					'@phpstan-assert',
+					new AssertTagValueNode(
+						new IdentifierTypeNode('Type'),
+						'$this',
+						false,
+						'assert Type to $this'
+					)
+				),
+			]),
+		];
+
+		yield [
+			'OK $this with generic type',
+			'/** @phpstan-assert GenericType<T> $this */',
+			new PhpDocNode([
+				new PhpDocTagNode(
+					'@phpstan-assert',
+					new AssertTagValueNode(
+						new GenericTypeNode(
+							new IdentifierTypeNode('GenericType'),
+							[
+								new IdentifierTypeNode('T'),
+							],
+							[
+								GenericTypeNode::VARIANCE_INVARIANT,
+							]
+						),
+						'$this',
+						false,
+						''
 					)
 				),
 			]),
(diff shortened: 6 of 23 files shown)


src/Ast/Type/ArrayShapeNode.php
```diff
@@ -13,7 +14,7 @@ class ArrayShapeNode implements TypeNode
 
 	use NodeAttributes;
 
-	/** @var ArrayShapeItemNode[] */
+	/** @var array<ArrayShapeItemNode|CommentNode|IdentifierTypeNode> */
```
ondrejmirtes: I don't think comments should be part of the AST as nodes. Being in node attributes would match how `nikic/PHP-Parser` works, and I would like that here too. It would make it easier to work with it, and also support comments everywhere, without BC breaks.

ondrejmirtes (changes requested):

src/Ast/Type/ArrayShapeNode.php
shmax: Okay, but can you please provide more of a big picture to me, here? ie:
1. why do we need to retain them at all?
2. does the printer need to reproduce them?
3. if so, what about new lines? Single line comments can depend on new lines to prevent them from affecting following content

And a little background about the printer would be helpful, because if that's what's producing the output of `dumpType` and what we see in error messages, then I imagine you would still want single-line output, in which case comments are just going to add noise, and if we agree on that then it's not clear to me why we shouldn't just ignore comments.

shmax: I took another crack at it:

* replaced simple // regex with a more robust pattern that captures single and block comments (so they can happen anywhere, as you say)
* modified `parseArrayShape` to catch comments. This seems to be necessary because array shape is one of the few node types that does recursive parsing (I had a look at nikic parser, by the way, and they have an architecture that works better for this kind of thing, with all the work of comment-capturing happening in a centralized `next` method that drives the entire process).

Am I getting warmer?

shmax:
I'm having some trouble negotiating with your CI build. I'm getting this error:

```
Time: 00:03.819, Memory: 80.00 MB

There was 1 PHPUnit test runner deprecation:

1) Your XML configuration validates against a deprecated schema. Migrate your XML configuration using "--migrate-configuration"!

--

There was 1 failure:

1) SlevomatCodingStandard\Sniffs\Commenting\ForbiddenAnnotationsSniffTest::testForbiddenAnnotations
Failed asserting that 9 is identical to 10.

/home/runner/work/phpdoc-parser/phpdoc-parser/slevomat-cs/tests/Sniffs/Commenting/ForbiddenAnnotationsSniffTest.php:24

FAILURES!
Tests: 718, Assertions: 2648, Failures: 1, Deprecations: 1.
```

So the tests are failing, but the reported error is some kind of slevomat thing? 😕

ondrejmirtes:
This is the workflow that fails in your PR: https://github.com/phpstan/phpdoc-parser/blob/1.23.x/.github/workflows/test-slevomat-coding-standard.yml

Slevomat CS is a heavy user of phpdoc-parser so we're testing every phpdoc-parser commit against Slevomat CS, to see that we didn't break anything.

The test in question:

* https://github.com/slevomat/coding-standard/blob/master/tests/Sniffs/Commenting/ForbiddenAnnotationsSniffTest.php
* https://github.com/slevomat/coding-standard/blob/master/tests/Sniffs/Commenting/data/forbiddenAnnotations.php

src/Ast/Comment.php
```diff
@@ -0,0 +1,24 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast;
+
+class Comment
+{
+
+	/** @var string */
+	protected $text;
```
ondrejmirtes: These should be public.

src/Ast/Attribute.php
```diff
@@ -13,4 +13,6 @@ final class Attribute
 
 	public const ORIGINAL_NODE = 'originalNode';
 
+	public const COMMENT = 'comment';
```
ondrejmirtes: This should be `COMMENTS` / `comments`.

ondrejmirtes (changes requested):
Please revert the commented-out tests, they shouldn't change or fail.

src/Ast/Type/ArrayShapeNode.php
ondrejmirtes: > why do we need to retain them at all?

It might be useful for the users of the parser to read them. For the same reason nikic/PHP-Parser retains them. And we might need it for the format-preserving printer.

> does the printer need to reproduce them?

You can write some tests for this in PrinterTest. If we remove an array shape item, its comment should also be removed. If we change something in an array shape item, its comment should be preserved.

You can also try to write a PrinterTest test case that removes/adds comments to AST nodes.

As for the Printer implementation, you can see the similarities between:

* https://github.com/nikic/PHP-Parser/blob/f4961b89ac0289455295e2dc2169527098a6e11b/lib/PhpParser/PrettyPrinterAbstract.php#L751-L998 (and also the `pArray` method)
* https://github.com/phpstan/phpdoc-parser/blob/9f854d275c2dbf84915a5c0ec9a2d17d2cd86b01/src/Printer/Printer.php#L506-L676

But you can also see all the places it touches node's comments there. I skipped these places when implementing the printer here because comments weren't a thing back then. But arguably we'll need to replicate the logic with comments here too, but the needs might not be that complex :)

> Am I getting warmer?

Yeah, probably :)

shmax:
> Slevomat CS is a heavy user of phpdoc-parser so we're testing every phpdoc-parser commit against Slevomat CS, to see that we didn't break anything.

Oh, that's clever, so you copy the current source of phpdoc-parser into slevomat's own vendor/phpstan folder, then you see what happens, eh? And because I'm now supporting comments within doc blocks, some test it runs related to "forbidden annotations" is getting throw off?

Any idea how to proceed, here?

src/Ast/Attribute.php
shmax: Fixed, thank you 829b5a0

src/Ast/Comment.php
shmax: 829b5a0

shmax:
An, yes, I had the tests commented out while I was experimenting with the node-based comment approach. Restored, thank you.

ondrejmirtes:
> Any idea how to proceed, here?

You should probably replicate what the workflow does locally, which will allow you to debug what's going wrong in the test. It might actually be valid, and we can just update the assertion in Slevomat CS after merging and releasing this PR. But it needs to be investigated.

shmax:
> You should probably replicate what the workflow does locally, which will allow you to debug what's going wrong in the test.

Great idea, thanks.

shmax:
Heh, well hurray for debugging. It turns out that the comment-parsing I was doing was getting thrown off by this kind of thing:

```
* @see https://www.slevomat.cz
```

I've modified my regex to include a negative lookbehind on ":", and that seemed to do the trick. Thanks much for the insights.

I guess that just leaves the format-preserving stuff, which I will look at this evening.

src/Ast/Type/ArrayShapeNode.php
shmax: I had a look through NIkic's printer; the section dealing with comments spans about 100 lines. As you say, it's vaguely similar to your code. I figure I'm looking at about a week or more of debugging and trying to understand what it means to add/remove shape items, and work out all the kinks. If the printer functionality is going to be a blocker on this PR then I guess I'll have no choice, but given that the tests pass as-is, is there any way we can leave that stuff for another time?

shmax: @ondrejmirtes Sorry to be a pest, but I'm blocked by this PR in my other work. Do you require the format-preserving printer stuff to move forward with this PR?

ondrejmirtes: Yes, it's a blocker, otherwise it'd mean I have to implement it myself after merging the PR without it.

This is a serious low-level library, I'd expect an issue here or in Slevomat CS pretty soon after people would start commenting their PHPDoc types.

shmax: Very well. I'll dive in on it, thanks much.

shmax: Made some progress. Still not sure I understand what I'm doing, but I've added a new printer test that uses the visitor pattern to add a few `ArrayShapeItems` with comments to a multiline array shape, and got the `DiffElem::TYPE_ADD` branch of `Printer::printArrayFormatPreserving` working, at least with single line comments. Whew.

Tomorrow I'll see what I can do about multi-line comments, but in the meantime can you shed any light on these CI errors?

```
[BC] ADDED: Method getComments() was added to interface PHPStan\PhpDocParser\Ast\Node
[BC] ADDED: Method getComments() was added to interface PHPStan\PhpDocParser\Ast\Type\TypeNode
[BC] ADDED: Method getComments() was added to interface PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocTagValueNode
[BC] ADDED: Method getComments() was added to interface PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocChildNode
[BC] ADDED: Method getComments() was added to interface PHPStan\PhpDocParser\Ast\ConstExpr\ConstExprNode
```

Thanks much.

ondrejmirtes: You can't add methods on interfaces, that's a BC break in case someone implements their own Node.

You always need to access the comments through getAttribute.

shmax: I see, thank you. Updated.

ondrejmirtes:
TBH I'd first make single line comments work perfectly,and only if there is a real demand add multiline comments. We're already inside a comment, we don't need our comments to support multiline comments 😀

shmax:
Fair point, but it was actually fairly easy to get working as all the hard work happens in `Comment::getReformattedText`, which I was able to lift unchanged from nickic. We'll see, though...

shmax:
With my latest changes, multiline comments seem to work with format-preserving printer:

```php
/**
 * @return array{
 *	c: string,
 *  /* This is a
 *     multiline
 *     comment
 *   */
 *  d: string,
 *  /* Another
 *   * multiline
 *   * comment with leading asterisks
 *   */
 *  e: string
 * }
 */
```

I also have the `DiffElem::TYPE_ADD` branch of `Printer::printArrayFormatPreserving` working.

I think that just leaves the `$delayedAdd` processing (which I haven't gotten as far as understanding, yet), and possibly removal of nodes, but I didn't see any removal stuff on array shapes already happening in the printer tests, so let me know if you want it.

shmax:
Oh, I also reverted the trailing comment style:

```php
/** 
 @return array{
    a: int // comment
  }
```

Nikic parser doesn't seem to support it, and I suppose it could arguably be considered ambiguous (does it go with the item to the left, or the item on the following line?), so I figure it's best to leave it for another day.

shmax:
Took a shot at implementing delayed add. At this point I think I'm ready for a round of reviews. I'm sure there are going to be a lot of use cases I missed, so let me know if you have any suggestions on how to improve the tests.

src/Lexer/Lexer.php
```diff
@@ -160,6 +163,7 @@ private function generateRegexp(): string
 			self::TOKEN_CLOSE_CURLY_BRACKET => '\\}',
 
 			self::TOKEN_COMMA => ',',
+			self::TOKEN_COMMENT => '((?<![:/])\/\/[^\n]*|\/\*(?!\*)[^*]*\*+([^\/][^*]*\*+)*\/)',
```
ondrejmirtes: How exactly this huge regex works? Is there a danger of it consuming more content than it should?

src/Parser/TypeParser.php
```diff
@@ -66,15 +75,29 @@ public function parse(TokenIterator $tokens): Ast\Type\TypeNode
...
 		}
 
+		if ($this->useCommentsAttributes) {
+			if ($tokens->currentTokenType() === Lexer::TOKEN_COMMENT) {
+				$comments[] = new Ast\Comment(
+					$tokens->currentTokenValue(),
+					$tokens->currentTokenLine(),
+					$tokens->currentTokenIndex()
+				);
+				$tokens->next();
```
ondrejmirtes: What if there's one comment after another? `/* c *//* d*/`

src/Parser/TypeParser.php
```diff
@@ -83,6 +106,18 @@ public function enrichWithAttributes(TokenIterator $tokens, Ast\Node $type, int
 		return $type;
 	}
 
+	/**
+	 * @param list<Ast\Comment> $comments
+	 */
+	public function appendComments(Ast\Node $type, array $comments = []): Ast\Node
```
ondrejmirtes: 1) should be private
2) I dislike optional parameters, make it required

src/Parser/TypeParser.php
```diff
@@ -761,10 +798,20 @@ private function parseArrayShape(TokenIterator $tokens, Ast\Type\TypeNode $type,
...
 			$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
-		} while ($tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA));
+			if (!$tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA)) {
+				$done = true;
+			}
+			if ($tokens->currentTokenType() !== Lexer::TOKEN_COMMENT) {
+				continue;
+			}
+
+			$this->appendComments($item, [new Ast\Comment($tokens->currentTokenValue(), $tokens->currentTokenLine(), $tokens->currentTokenIndex())]);
```
ondrejmirtes: It's a pity that this PR adds comments only specifically to array shape items. It should be achieved somehow generally so that all types that support multiline syntax can support comments automatically.

These types come to mind:

1) Object shapes (very similar but there are different methods for them in TypeParser)
2) Conditional types
3) Union and intersection types
4) Generic types
5) Callables

Again, it should be possible to achieve this automatically without modifying parsing each of these types which can be very error-prone to implement.

I see that you added reading a comment in `enrichWithAttributes`. What happens, what kind of syntax support we lose, if we just leave the modification in enrichWithAttributes, and remove every change about parsing array shapes?

ondrejmirtes (changes requested):

src/Lexer/Lexer.php
shmax: Explanation:

1. `((?<![:/])\/\/[^\n]*|\/\*[\s\S]*?\*\/)`: This entire regular expression is composed of two main alternatives separated by a pipe |.

1. First alternative: `(?<![:/])\/\/[^\n]*`

* `(?<![:/])`: Negative lookbehind to ensure that `//` is not preceded by `:` or `/`, thus avoiding URLs.
* `\/\/`: Matches the start of a single-line comment `//`.
* `[^\n]*`: Matches any character except a newline `\n`, effectively matching the rest of the comment line until the end.


3. Second alternative: `\/\*[\s\S]*?\*\/`

* `\/\*`: Matches the start of a multi-line comment `/*`.
* `[\s\S]*?`: Non-greedy match for any character, including newlines, zero or more times. This matches the content of the multi-line comment.
* `\*\/`: Matches the end of a multi-line comment `*/`.

> Is there a danger of it consuming more content than it should?

I don't think so, but I guess we won't know for sure until it happens. I'm available to support if it comes up.

src/Parser/TypeParser.php
shmax: fair enough: 5d9be65

Although I don't really know what it means, or how this all gets consumed by other libraries.

src/Parser/TypeParser.php
shmax: Absolutely. cee33d9

src/Parser/TypeParser.php
shmax: I can try, but I think it's complicated by architectural differences between this library and `PHP-Parser`:

`PHP-Parser` has a central `getNextToken` hook in `Lexer`, and everything must go through it to get tokens. Comments are  plucked from the token list and appended to an attributes structure that is pushed onto a stack, and used to decorate nodes later.

`phpdoc-parser` requires us to manually call `enrichWithAttributes` over and over. 

Anyway, let me poke around a bit more. Maybe we can do something.

shmax: > and remove every change about parsing array shapes? 

I don't think we can really do that part of it. I just tried it, and the parser travels from `parseArrayShape` to `parseArrayShapeItem`, and then finally crashes in `parseArrayShapeKey`. I could maybe adjust `parseArrayShapeKey` to not die on comments, but by that point we've already gone wrong, because if I understand correctly we want the comment to be associated with the `Ast\Type\ArrayShapeItemNode` node, and not the key node.

ondrejmirtes: I'm not against rearchitecting this library (in a separate PR) to make our job easier here in this PR.

shmax: I think that could help with general comment parsing, but for the particular case of array shape items I think we're going to have to continue special-casing (because we want to pair comments specifically with items,  and not keys or values).

But if you want to start thinking about how to remove the need for all those calls to `enrichWithAttributes` it could help with general comment parsing.

ondrejmirtes: > But if you want to start thinking about how to remove the need for all those calls to enrichWithAttributes

That would be an interesting exercise! If you have ideas about this please send them in a separate PR 😊

shmax: Well, I walked right into that one. 😆

shmax: I'm going to try to make the same changes for object shape, but before I do that I thought it might help to make the printer test cases easier to read. In one of my other projects I made a little multi-line helper called `nowdoc`, and I added it here:

05e9eedc3c9cf58ff0a8c0cf2dbc76501222f3f1

Any objections to something like this? If so, I'll revert it, if not, I'll continue.

shmax: Object comments seem to work:

```php
yield [
			self::nowdoc('
				/**
				 * @return object{
				 *   // your favorite bar
				 *   bar: string
				 * }
				 */'),
			self::nowdoc('
				/**
				 * @return object{
				 *   // your favorite bar
				 *   bar: string,
				 *	 // favorite foo
				 *	 foo: int
				 * }
				 */'),
			$addItemsWithCommentsToObjectShape,
		];
```

shmax: > These types come to mind:

> Object shapes (very similar but there are different methods for them in TypeParser)
> Conditional types
> Union and intersection types
> Generic types
> Callables

Can you give an example of what each of these would look like with a comment? Don't worry, I don't intend to special-case each one, but I would like to know what you envision.

ondrejmirtes:
I've got an idea - every `$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);` in TypeParser should be replaced by a special method on TokenIterator that would return an array of comments which we could use in enrichWithAttributes. Because comments only make sense at the end of the line, it makes sense to pair it with consuming EOL.

src/Parser/TypeParser.php
ondrejmirtes: Imagine multi-line variants of these types, and imagine there's `// something` at the end of each line.

shmax:
> I've got an idea - every $tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL); in TypeParser should be replaced by a special method on TokenIterator that would return an array of comments which we could use in enrichWithAttributes. Because comments only make sense at the end of the line, it makes sense to pair it with consuming EOL.

Sure, can try something like that, but that's one step removed from just doing whatever you're going to do directly in `enrichWithAttributes`, isn't it? And I'm not sure that a comment is always at the end of the line, for example:

```
/*****************************
 * @param /* TODO: remove */  array $data
 */
```

src/Parser/TypeParser.php
shmax: I'll have to experiment with Nikic parser a bit more, but I'm a little unclear if above-the-line and end-of-the-line pair a comment with the same item:

```php
array{
  // a is for apple
 a: string,
b: int
}
```

vs:

```php
array{
a: string, // a is for apple
b: int
}
```

Are these both annotating `a: string` with the same comment?

shmax: I ran with your idea of having `TokenIterator` do more of the work, and managed to remove all the special comment handling from the array/object shape stuff, and got things working in a general sense: [0558a8b](https://github.com/phpstan/phpdoc-parser/pull/213/commits/0558a8b1341d01eb56a9c4d365f5714363aae968)

The only problem with adding comments to things like Generics and Conditionals is that they aren't written to handle multiple new lines, so I had to adjust their parsing routines to loop over new lines. 

Let me know if I'm on the right track, here, and I'll see if I can finish up.

shmax:
Moving this back to draft, as I'm discovering general problems around format-preserving of single line vs inline comments. Hopefully won't take too long to sort out...

shmax:
> general problems around format-preserving of single line

False alarm. I confirmed behavior against Nikic parser, and the behavior with comments in this branch seems to be correct. I had the idea that inline comments should stay inline when format-preserving printing, but apparently not; all comment types force a newline after a comment, and then another after the item it is paired with.

src/Parser/TypeParser.php
shmax: > intersection types

I tried to get  intersections working with comments, but multiline syntax doesn't seem to work; this doesn't work with or without comments:

```php

/**
 * @return 
 *    Foo
 *      &
 *    Bar
 * /
```

I can wok on that if you really want it, but otherwise I'd be much happier to leave it for another time.

shmax: > Conditional types

This also won't work with comments and format-preserving printer in its current form, because it doesn't have a single `items` property of array type, so it will either need to be refactored to work more like `ArrayShapeType` and `ObjectShapeType`, or I will need to do special handling in the printer. I'm always interested in the challenge, but would love to save this for another day if you don't mind.

shmax: > Callables

This one (nearly--had to eat extra lines) works out of the box:

[b00d20a](https://github.com/phpstan/phpdoc-parser/pull/213/commits/b00d20ac96a0ab94cc22f409c97d6519781fbde0)

shmax: > Union

This one is nearly doable, but the code in Nikic parser seems to forbid it:

```php
// We go multiline if the original code was multiline,
                // or if it's an array item with a comment above it.
                if ($insertStr === ', ' &&
                    ($this->isMultiline($origNodes) || $arrItem->getComments())
                ) {
                    $insertStr = ',';
                    $insertNewline = true;
                }
```

So, nikic parser only prints comments when it's a newline, and we only decide to do a new line when there are comments, or the insert string is `,`, but for union the insert string is `|`. I can tweak the logic if you want, but I think it will make us diverge from nikic parser.

At this point, I think I have done just about everything I can. Please advise.

shmax: > Are these both annotating a: string with the same comment?

Finally, I did more debugging with Nikic parser, and no, it seems that a comment that comes after an array item is _not_ paired with the item that precedes it; it is associated with the following line:

```php
$items = [
    'a' => 'foo',  // puppies are cute
    'b' => 'bar'
];
```

In this example, the "puppies" comment is bound to the `'b' => 'bar'` line.

ondrejmirtes:
About `@param /* TODO: remove */  array $data` - I don't think we want to support this because the comment is not part of the type, it precedes it. And from this it's also obvious why I want to start by supporting end-line `// comment` only :)

src/Parser/TokenIterator.php
```diff
@@ -180,16 +184,33 @@ public function tryConsumeTokenValue(string $tokenValue): bool
...
+	{
+		$res = $this->comments;
+		$this->comments = [];
+		return $res;
+	}
+
 
 	/** @phpstan-impure */
-	public function tryConsumeTokenType(int $tokenType): bool
+	public function tryConsumeTokenType(int $tokenType, bool $consumeAll = false): bool
```
ondrejmirtes: I really don't like `bool` arguments and I really don't like optional arguments. Please make the scenario where you'd call this with `true` as a separate method with a different name. Thank you.

ondrejmirtes (changes requested):

src/Parser/TokenIterator.php
shmax: Fair enough, let me revisit this.

shmax: Better? 1c3f514

shmax:
> About @param /* TODO: remove */  array $data - I don't think we want to support this because the comment is not part of the type, it precedes it. And from this it's also obvious why I want to start by supporting end-line // comment only :)

I'll do whatever you want to move things forward, but the sample we're discussing here doesn't really have an impact on the array-based stuff I've been working on (I've confirmed over and over that those are working perfectly in alignment with nikic parser, for both single-line and block comments).

I'm not sure what you expect to happen for  `/* TODO: remove */ $array` sample, but from my testing it does indeed get bound to the following item, in this case the identifier. The comment doesn't break anything, but the format-preserving printer doesn't currently know how to handle these kind of comments. I can continue to investigate that if you like...

edit: to clarify, such a comment _does_ appear in the printer output, but not if you "edit" the identifier node in this use case to include a new comment. I'm having trouble finding an equivalent sample in the nikic parser tests to analyze, but its format-preserving printer has a few branches that yours doesn't (eg. `pStmts`).

ondrejmirtes:
`/* ... */` inside `/**` PHPDoc is not even valid PHP: https://3v4l.org/R7jBd

So please remove that from the parser along with its complexity and leave only the trailing `//` comment. Thank you.

shmax:
Well, that's certainly a fair point! I don't think it will reduce complexity at all, but obviously if it breaks PHP it will have to go. Let me see what I can do...

shmax:
Here you go: bfa599138b9a097073663319cb17a31d4e7f402e

ondrejmirtes:
What about all of this?

Screenshot 2023-10-05 at 15 53 42

shmax:
Sure: 3b6ff9a

src/Parser/TokenIterator.php
```diff
@@ -180,12 +184,30 @@ public function tryConsumeTokenValue(string $tokenValue): bool
...
-		$this->skipIrrelevantTokens();
+		$this->next();
 
 		return true;
 	}
 
+	/**
+	 * @return Comment[]
+	 */
+	public function flushComments(): array
```
ondrejmirtes, line 195: i think this PR is starting to be in a pretty good shape. I still have to go through the many test cases and think about it very hard (and try to break your logic :))

Every time there's a new stateful logic introduced like here with the comments array, I try to think how we can prevent it being used wrongly.

I think TokenIterator should throw an exception if the comments weren't flushed (the array is not empty):

1) Definitely when consuming TOKEN_END
2) Maybe even more often, basically every time we assume the array should be empty. Like when consuming some other token than EOL and maybe others.

ondrejmirtes (changes requested):

src/Parser/TokenIterator.php
shmax: I'm having a look through the code, too, but so far I don't actually see any calls to `consumeTokenType` with `TOKEN_END` as an argument.

My next thought was to do a check before returning from `TypeParser::parse`, but that one is recursive, so `TypeParser` doesn't really know when it's done (without some further refactoring).

`PhpDocParser::parse` has some potential, but while it is possible to have comments remaining when the parsing is done, it's not clear to me if it's an error. For example, this seems like it should be valid to me:

```php

/**
 *  @return int|false
 *
 * // can return false on failure
 * /

```

Since we always attach each comment to the following node, and there are no more nodes after the comment, we technically have a leftover comment, but no cause to throw an exception.

ondrejmirtes: Alright, so this comment made me hear a loud sound of breaking tableware :D

We need to make sure that comments outside of types, outside of TypeParser, are still parsed as they are today!

So for example:

```php
/** @param int $a // this is a description */
```

Must become a description of ParamTagValueNode. Please be aware that there are two differents implementations of consuming descriptions: `parseOptionalDescription` and `parseOptionalDescriptionAfterDoctrineTag`. So make sure to update and test both. Thank you.

In this case:

```php
/**
 *  @return int|false
 *
 * // can return false on failure
 * /
```

This must become a standalone PhpDocTextNode as today.

Can you make sure they still do?

shmax: > Can you make sure they still do?

~~Did they ever? I didn't know about `parseOptionalDescription` or `parseOptionalDescriptionAfterDoctrineTag`, but I checked out the `1.23.x` branch, stuck a breakpoint in each of them, and ran the tests. Neither breakpoint was hit.~~

edit: disregard. Somehow I wound up inside the `build-cs` folder.

Can you provide a breaking test for me? Feel free to push a commit.

shmax: > /** @param int $a // this is a description */
> Must become a description of ParamTagValueNode

By the way, I'm not sure about that one. None of the samples I see in the tests look like that. They all look like this:

```php
/** 
@param int $a this is a description 
*/
```

ondrejmirtes: Here it is: https://github.com/phpstan/phpdoc-parser/commit/0bb2fe4cc43a6144d3135facec4d036258a5e4dd

Please squash and rebase your branch so that the history is linear and clean. I was unable to do so.

tests/PHPStan/Parser/PhpDocParserTest.php
```diff
@@ -5526,6 +5562,98 @@ public function provideSelfOutTagsData(): Iterator
 		];
 	}
 
+	public function provideCommentLikeDescriptions(): Iterator
+	{
+		yield [
+			'Comment after @param',
+			'/** @param int $a // this is a description */',
```
shmax, line 5569: It is *not* a description, it is a _commented_ description, meaning that as far as the parser is concerned there is no description here.

tests/PHPStan/Parser/PhpDocParserTest.php
```diff
@@ -5526,6 +5562,98 @@ public function provideSelfOutTagsData(): Iterator
...
+					'// this is a description'
+				)),
+			]),
+		];
+
+		yield [
+			'Comment on a separate line',
+			'/**' . PHP_EOL .
+			' * @param int $a' . PHP_EOL .
+			' * // this is a comment' . PHP_EOL .
```
shmax, line 5584: It's not a text node. It is a commented text node., meaning that there is no tag or descriptive content here.

src/Parser/TokenIterator.php
shmax: Wonderful, thank you. I've rebased everything down to just our two commits so we can get a better look at things.

As much as I hate to disagree with you, I feel I must question your premise. I feel that you are getting comments and descriptions mixed-up. A description is text that comes after a tag value, and a comment denotes text as having no functional meaning. So, a commented description is _not_ a description, it's simply a comment. The same goes for a comment sitting on a line by itself.

I've spent many hours debugging nikic parser, and what I've learned is that comments are always attached to the following token, and never the preceding one.

So, in order to achieve what you are asking I would need to violate both the fundamental functionality of the parser this one is modeled after, and also the principle of least surprise, in that we are stripping the "do nothing" nature of comments.

That said, I still don't know how nikic parser handles comments that come at the end of a PHP file and have no tokens to attach to, but if you will bear with me I will find out.

shmax: > how nikic parser handles comments that come at the end of a PHP file

The answer seems to be: it does nothing with them.

They are parsed, a new `Comment` instance is created, and it is added to the running `startAttributes` object kept around during execution of the parse algorithm, but it is not ultimately added to any nodes.

shmax: Maybe this will help illustrate how I'm viewing the difference. Compare:

```php
/**
 * @param int $foo the number of foos to count
 */
 ```

vs

```php
/**
 * @param int $foo // TODO: come up with description
 */
 ```

Comment (shmax) [reply src/Parser/TokenIterator.php:195]: @ondrejmirtes please let me know your thoughts.

Comment (ondrejmirtes) [reply src/Parser/TokenIterator.php:195]: I understand this effort is to make comments inside types possible.

Otherwise we're already inside a comment. There's no structural difference between `@param int $foo the number of foos to count` and `@param int $foo // TODO: come up with description`.

`//` outside of PHPDoc types must mean the same thing as before. If not for nothing else then at least for backward compatibility. I don't want to change meaning of something that's already possible. I just want to make a new feature possible that's currently a syntax error.

Comment (shmax) [reply src/Parser/TokenIterator.php:195]: > Otherwise we're already inside a comment. There's no structural difference between @param int $foo the number of foos to count and @param int $foo // TODO: come up with description.

There is a functional difference. Let's rewrite it a different way:

```php
// with description
$tag = [
  'name' => '@param',
  'type' => `int`,
  'parameterName' => '$foo,
 'description` => 'the number of foos to count'
];
```

```php
// with comment
$tag = [
  'name' => '@param',
  'type' => `int`,
  'parameterName' => '$foo,
 'description` => '' // TODO: come up with a description
];

```

Is that more clear? The first one has a description, the second one doesn't.

> // outside of PHPDoc types must mean the same thing as before. If not for nothing else then at least for backward compatibility.

Can you elaborate on this? I'm not clear on what you feel is broken. Comments are currently ignored, which is what is supposed to happen. How exactly have I broken backwards-compatibility? All the existing tests pass, and comments are neatly ignored. 

If you want to add the kind of complexity you have in mind later that's your business, but for now do we really have a problem?!

ondrejmirtes: About backward compatibility - I've described the current behaviour which we need to keep in these tests: https://github.com/phpstan/phpdoc-parser/commit/0bb2fe4cc43a6144d3135facec4d036258a5e4dd

AFAIK they are failing in your PR.

shmax: > How exactly have I broken backwards-compatibility?

Ah, maybe I understand. You're saying that there is code currently floating around out in the wild where there are already comments assigned to descriptions and being interpreted as descriptions, and your concern is that after this change they will lose their description status, is that right?

ondrejmirtes: Yes, we can't change the AST like that. It'd also make altering these comments more difficult and diferent for the users of format-preserving printer, most notably Slevomat CS.

shmax: Okay. At this point I am going to abandon this PR. I tried. I feel that the behavior in this PR is correct. I understand it may "break" descriptions out in the wild, but I would argue that the code out in the wild is incorrect, and this fixes it. That happens sometimes in the software development cycle, and the solution is to do a major version release, not add tons of complexity trying to achieve backwards compatibility with existing behavior that is wrong.

I'm going to use my own fork, which I really hate to do, but I've done everything I can within reason to solve this problem rationally, and I'm not getting support. Good luck.

ondrejmirtes:
I'm sorry that you feel this way and that you're not going to finish this PR. But I need to do what I feel is right for this project and its users, and also to keep the support burden low and maintainability high.

shmax:
Do whatever you want, dude.

shmax:
Sigh. I forgot I can't just fork the parser, because you have everything all wrapped up in a phar.

ethernidee:
@shmax, multiline comments would be really great feature. We are grateful to you for all the efforts.

shmax:
@ethernidee thanks, I appreciate the words of encouragement. Let me summarize the situation:

1. I opened a PR with a very simple solution, which was to ignore array shape comments
2. @ondrejmirtes decided that if we are going to have comments, then we must do them the same way that nikic parser does them, and that comments must work everywhere, not just in array shape. 
3. I got that working, and now @ondrejmirtes has declared that they should NOT work as one would expect when used in lines with tag names, because that would not be "backwards compatible"

I can't spent the rest of my life trying to hit a moving target, so I decided to just document my own code the old-fashioned way (mkdocs), but I may return to this at some point if I get bored. In the meantime, feel free to take this code and open your own PR.

ethernidee:
To my mind, multiline comments are necessary everywhere. Array shape, object shape, regular tags. It's normal practice to be able to describe something using ascii-graphics, a few sentences or some kind of a list. The same way, as regular comments work.

shmax:
>  multiline comments are necessary everywhere

Unfortunately, we've discovered that multiline comments aren't legal inside a doc block comment (PHP can't parse them).

stof:
At least, you would have to use a different delimiter for the multiline comments that the PHP delimiter. Otherwise, the `*/` would end the PHP comment (and then, the next content would be PHP code, not the next part of the phpdoc)

alexandre-daubois:
@shmax Is there something we can do to help this PR going forward? 🙂

ondrejmirtes:
I'm going to take over and finish it (I already talked to Nicolas about it), it shouldn't be too much work.

shmax:
Oh, yeah, sorry guys. I'm glad @ondrejmirtes  is going to take over on this, but as for me I've finally had it with PHP. I hung in there for 20 years, but after working with TypeScript for the last decade (and more recently NodeJs) I've come to the conclusion that I really can't live without the power of TypeScript's interface type system (PHPstan's array shapes may catch up someday, but there doesn't seem to be any movement).

Best of luck.

ondrejmirtes:
@shmax Feel free to open any feature requests that would make sense for PHP & PHPStan. Outside perspective from users of other languages is welcome.

BTW Some people feel the opposite way https://x.com/SlyFireFox/status/1870877014284173464.

ondrejmirtes:
Superseded by https://github.com/phpstan/phpdoc-parser/pull/264

Issue #212, opened 2023-09-17
PHPStan report a Exception when upgrade to v1.10.34


rainx:
# What's happening

When I was trying to submit a PR https://github.com/phpstan/phpdoc-parser/pull/211 without any functional code update. the CI failed with the PHPStan job which should have run correctly before.

Compare with two CI workflows(before and after) 
- After(failed):  https://github.com/phpstan/phpdoc-parser/actions/runs/6206864021/job/16851520190?pr=211 
- Before(successful): https://github.com/phpstan/phpdoc-parser/actions/runs/6144706750/job/16670656414

we found the main difference was for the installed dependencies. 
and the PHPStan just updated to v1.10.34 for the latest PR's CI job.

# How to fix
1. Lock the dependency version by adding the lock file to the repo. and let Renovate handle the dependencies updating. It assures every developer can get the same environment at different moments.
2. Fix the PHPStan-reported issue in the main branch

rainx:
Here is the output that PHPStan report in the workflow 
```sh
php vendor/bin/phpstan
Note: Using configuration file /-----/---/phpdoc-parser/phpstan.neon.
 103/103 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ------------------------------------------------------------------------------------------------------------------- 
  Line   src/Ast/NodeTraverser.php                                                                                          
 ------ ------------------------------------------------------------------------------------------------------------------- 
         Ignored error pattern #^Strict comparison using \=\=\= between 2 and 2 will always evaluate to true\.$# in path    
          /-----/---/OpenSourceProjects/other/phpdoc-parser/src/Ast/NodeTraverser.php was not matched in reported errors.  
 ------ ------------------------------------------------------------------------------------------------------------------- 
```

rainx:
To work around this. I submit a new commit in PR #211 https://github.com/phpstan/phpdoc-parser/pull/211/commits/0f4141384bba77347690d6a5f0c56fb868f986a3

ondrejmirtes:
Please send that fix as a separate PR, I'm gonna smash that green button 😊

rainx:
@ondrejmirtes done! check it out https://github.com/phpstan/phpdoc-parser/pull/214

PR #214, opened 2023-09-18
Make the CI job pass after upgrading PHPStan to v1.10.34

rainx:
This PR is to fix the issue https://github.com/phpstan/phpdoc-parser/issues/212


ondrejmirtes:
Thank you!

PR #215, opened 2023-09-18
Deduplicate generics parsing

mvorisek:
extracted from #199

previously the returned type was build on two places, newly, the code is deduplicated


ondrejmirtes:
Thank you!

PR #217, opened 2023-09-26
Allow conditional type in callable return type

### LINKED ISSUE #151: Using `class-string<T>` in a conditional `Closure` return type results in parsing error.
# Bug report

Using `class-string<T>` in a conditional `Closure` return type results in parsing error.

```
PHPDoc tag @param has invalid value (Closure(Container):($serviceId is class-string<TService> ? TService : mixed)) $service): Unexpected token "(", expected variable at offset 150    
```

```php
<?php

namespace Phpactor\Container;

use Closure;

interface ContainerBuilder
{
    /**
     * @template TService
     * @param class-string<TService>|string $serviceId
     * @param Closure(Container):($serviceId is class-string<TService> ? TService : mixed) $service
     */
    public function register(string $serviceId, Closure $service): void;
}
```

https://phpstan.org/r/17633639-1608-478a-b439-6fc15373daf5

### Did PHPStan help you today? Did it make you happy in any way?

PHPStan helped me several times today, it's awesome.

rvanvelzen:
Fix #151

### DIFF
--- a/src/Parser/TypeParser.php
+++ b/src/Parser/TypeParser.php
@@ -524,7 +524,7 @@ class TypeParser
 			return $this->parseNullable($tokens);
 
 		} elseif ($tokens->tryConsumeTokenType(Lexer::TOKEN_OPEN_PARENTHESES)) {
-			$type = $this->parse($tokens);
+			$type = $this->subParse($tokens);
 			$tokens->consumeTokenType(Lexer::TOKEN_CLOSE_PARENTHESES);
 			if ($tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_SQUARE_BRACKET)) {
 				$type = $this->tryParseArrayOrOffsetAccess($tokens, $type);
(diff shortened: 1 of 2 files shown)


ondrejmirtes (changes requested):
Every parser change now has implications for the Printer.

This change now means that the return type can contain NullableTypeNode which has to be parenthesized.

This change should be tested by trying to print CallableTypeNode with NullableTypeNode in the return type. The change needs to happen here: https://github.com/phpstan/phpdoc-parser/blob/9f854d275c2dbf84915a5c0ec9a2d17d2cd86b01/src/Printer/Printer.php#L368

Also the format-preserving printer needs an addition in `Printer::$parenthesesMap`: https://github.com/phpstan/phpdoc-parser/blob/9f854d275c2dbf84915a5c0ec9a2d17d2cd86b01/src/Printer/Printer.php#L122-L126

rvanvelzen:
Nullable types were already accepted, and do not need to be parenthesized. The only change is that conditional types are now accepted.

ondrejmirtes:
Oh right, I didn't realize that.

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #218, opened 2023-09-28
Add # as vallid character for variable.

Blurazzle:
This pull request is related to the issue https://github.com/phpstan/phpstan/issues/9919#issuecomment-1733364261. The problem that it resolves can be seen in https://phpstan.org/r/459a7de6-aa15-42ee-8967-0777088f411e.

Our application is connected to an IBM i DB2 database. The column names are sometimes like the following
 * @property string $PBGAF#
 * @property string $PBGTL#
 * @property string $PBGD#L

Currently, the lexer does not recognize these as valid property names, while in fact, these are valid properties in PHP that can be accessed with $this->{'PBGD#L'} syntax.


ondrejmirtes:
We can't do this change so lightly. It's a BC break and it's questionable whether we even want that.

PR #220, opened 2023-11-17
fix HTML description detection for phpstorm stubs

schlndh:
Here is a quick attempt to address [this](https://github.com/phpstan/phpstan-src/pull/2622#issuecomment-1816590052)

The problem turned out to be due to the fact that the existing code relies on detecting the `<` of the end tag as a separate token first. But instead it gets `.</p>` as a single token.

I'm not sure whether it's better to fix it this way or by modifying the lexer.

BTW: README.md says that you can just run `composer install` and `make` to build it, but `make cs` doesn't work (build-cs directory is missing - github actions check it out from phpstan's repo).


ondrejmirtes:
Thank you!

PR #223, opened 2023-11-26
Git export-ignore phpstan-baseline.neon

herndlm:
just a drive-by PR because I saw this file in my vendor


ondrejmirtes:
Thank you.

PR #226, opened 2024-01-04
Added support for @require-extends phpdoc

staabm:


staabm:
should I do `@require-implements` in this very same PR?

edit: just realized I am missing the "printing" parts

src/Ast/PhpDoc/PhpDocNode.php
```diff
@@ -187,6 +187,19 @@ static function (PhpDocTagValueNode $value): bool {
 		);
 	}
 
+	/**
+	 * @return RequireExtendsTagValueNode[]
+	 */
+	public function getRequireExtendsTagValues(string $tagName = '@require-extends'): array
```
ondrejmirtes: Similar tag has a prefix here by default: https://github.com/phpstan/phpdoc-parser/blob/a3b405864280bbae461c89e2e6a47ffee3fb86c9/src/Ast/PhpDoc/PhpDocNode.php#L291 (We don't need to support and test prefix-less variant).

ondrejmirtes (changes requested):
Yes, please add `@phpstan-require-implements` here too.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #230, opened 2024-02-07
Improve static keyword conflict resolution in @method [closes #229]

### LINKED ISSUE #229: Return type in `@method` starting with `static` cannot be parsed
### Bug report

I know the current workaround is to place parenthesis around the return type, but this is problematic, as almost every CS fixer removes the parenthesis.

Like `\Closure(): void`, ie. type even with space is supported [1], unioned (and intersected) types should be supported as well.

[1] https://phpstan.org/r/e9b294c3-2f5f-42a1-b4a7-c2ad93a83c2f

https://phpstan.org/r/ad8c4bfc-4c57-48fc-838e-fc3d14a6d491

### Expected output

no phpstan parser error

JanTvrdik:


tests/PHPStan/Parser/PhpDocParserTest.php
```diff
@@ -2626,6 +2626,26 @@ public function provideMethodTagsData(): Iterator
 				),
 			]),
 		];
+
+		yield [
+			'OK non-static with return type that starts with static type',
+			'/** @method static|null foo() */',
```
mvorisek, line 2632: would be also good to test `@method static |null foo()`

tests/PHPStan/Parser/PhpDocParserTest.php
```diff
@@ -2626,6 +2626,26 @@ public function provideMethodTagsData(): Iterator
...
+			new PhpDocNode([
```
mvorisek, line 2633: any idea about https://github.com/phpstan/phpdoc-parser/issues/229#issuecomment-1932810377 fuzzer testing?

mvorisek:
@ondrejmirtes can this fix be merged?

ondrejmirtes:
Thank you!

PR #232, opened 2024-02-23
Parse generic callable

mad-briller:
Follow on from: https://github.com/phpstan/phpdoc-parser/pull/199 originally by @mvorisek 

~Making `TemplateTagParser::parseTemplateTagValue()` static seemed like the best approach to fix a circular dependency between `TypeParser` and `TemplateTagParser`, as it means the `TypeParser` must be passed for each call.~

Also `parseOptionalDescription` was awkward to handle as it only needs to be used in `PhpDocParser` and calls many private functions of `PhpDocParser`, so i reworked that to be a callable instead of a boolean flag.


src/Parser/TemplateTagParser.php
```diff
@@ -0,0 +1,47 @@
...
+use PHPStan\PhpDocParser\Lexer\Lexer;
+
+class TemplateTagParser
+{
+
+	/**
+	 * @throws ParserException
+	 * @param ?callable(TokenIterator): string $parseDescription
+	 */
+	public static function parseTemplateTagValue(
```
ondrejmirtes: This can be a public instance method in TypeParser, right? That way it can be called from TypeParser and from PhpDocParser (which already contains TypeParser).

ondrejmirtes (changes requested):

src/Parser/TemplateTagParser.php
mad-briller: you are right, it can :sweat_smile: got tunnel visioned by the comment on the old pr

tests/PHPStan/Parser/TypeParserTest.php
```diff
@@ -988,11 +988,11 @@ public function provideParseData(): array
 						),
 					]),
 					[
-						new CallableTypeTemplateNode(new IdentifierTypeNode('Tx'), new UnionTypeNode([
+						new TemplateTagValueNode('Tx', new UnionTypeNode([
 							new IdentifierTypeNode('X'),
 							new IdentifierTypeNode('Z'),
-						])),
-						new CallableTypeTemplateNode(new IdentifierTypeNode('Ty'), new IdentifierTypeNode('Y')),
+						]), ''),
```
mvorisek, line 994: https://github.com/phpstan/phpdoc-parser/blob/1.25.0/src/Ast/PhpDoc/TemplateTagValueNode.php#L26 `description` should be made optional

tests/PHPStan/Parser/TypeParserTest.php
```diff
@@ -913,7 +913,7 @@ public function provideParseData(): array
 					],
 					new IdentifierTypeNode('C'),
 					[
-						new CallableTypeTemplateNode(new IdentifierTypeNode('A'), null),
+						new TemplateTagValueNode('A', null, ''),
```
mvorisek, line 916: I would prefer to change https://github.com/phpstan/phpdoc-parser/blob/1.25.0/src/Ast/PhpDoc/TemplateTagValueNode.php#L15 first to keep `IdentifierTypeNode` instead of `string` to allow to store enriched attributes like position in the source phpdoc.

mad-briller: changing `TemplateTagValueNode` would be a BC break though, and any other solution would require not reusing what already exists

src/Ast/Type/CallableTypeNode.php
```diff
@@ -13,6 +14,9 @@ class CallableTypeNode implements TypeNode
 	/** @var IdentifierTypeNode */
 	public $identifier;
 
+	/** @var TemplateTagValueNode[] */
+	public $templates;
```
ondrejmirtes: In `MethodTagValueNode` this is called `$templateTypes`

ondrejmirtes (changes requested):
This looks really solid! 👍

ondrejmirtes:
Perfect, thank you!

PR #233, opened 2024-03-13
Upgrade to `nikic/php-parser:^5.0`

tanerkay:
see https://github.com/nikic/PHP-Parser/blob/master/UPGRADE-5.0.md

notably a php 7.4 version requirement

if we want to retain php 7.2/7.3 support, we could support both v4.0 and v5.0, but given that the main phpstan repo already requires php 8.1, I think this is fine, given the age of php 7.3 (end of life since over 2 years)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,8 +16,6 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.2"
-          - "7.3"
           - "7.4"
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
phpdoc-parser does not depend on nikic/PHP-Parser. It's only in require-dev.

If you're getting an error message about a conflict in Composer, it's not a fault of phpdoc-parser.

PR #234, opened 2024-03-23
Allow for v5 of nikic/php-parser

adrum:


ondrejmirtes:
See https://github.com/phpstan/phpdoc-parser/pull/233

PR #235, opened 2024-04-03
Support `@phan-` prefixes for recognized doc tags

anomiex:
The following prefixes are added:

* `@phan-assert`
* `@phan-assert-if-false`
* `@phan-assert-if-true`
* `@phan-extends`
* `@phan-inherits`
* `@phan-method`
* `@phan-mixin`
* `@phan-param`
* `@phan-property`
* `@phan-property-read`
* `@phan-property-write`
* `@phan-real-return`
* `@phan-return`
* `@phan-template`
* `@phan-type`
* `@phan-var`

No changes to any of the parsing were made, as the syntaxes seem to match what's already done for the existing unprefixed tags and/or the existing `@psalm-` and `@phpstan-` prefixed tags.

### DIFF
--- a/src/Parser/PhpDocParser.php
+++ b/src/Parser/PhpDocParser.php
@@ -384,6 +384,7 @@ class PhpDocParser
 				case '@param':
 				case '@phpstan-param':
 				case '@psalm-param':
+				case '@phan-param':
 					$tagValue = $this->parseParamTagValue($tokens);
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you!

ondrejmirtes:
Please send a PR to phpstan-src too.

PR #237, opened 2024-05-06
Support array and offset access on const types

rvanvelzen:
See https://github.com/phpstan/phpstan-src/pull/3050#issuecomment-2095611108 for context


ondrejmirtes:
Added some fixes https://github.com/phpstan/phpdoc-parser/pull/237/commits/536451b77bd7fcee653144520d3bf1b52b836c24

ondrejmirtes:
Thank you!

Issue #173, opened 2022-12-22
"New in initializers" breaks in docblock


juampi92:
# Bug report

Since PHP 8.1, it is possible to use `new` as default parameters (["new in initializers"](https://www.php.net/releases/8.1/en.php#new_in_initializers) [RFC](https://wiki.php.net/rfc/new_in_initializers)).

PHPStan gets it perfectly, but things stop working once we move them to docblocks, explicitly the `@method` docblock (see example).

### Code snippet that reproduces the problem

https://phpstan.org/r/5d0cdcbb-5c11-4483-90b0-b95986f92bc2

### Expected output

Since PHPStan understands the "new in initializers" in the code, I guess that this is a missed use-case instead of an actual bug, but I didn't know where to report it.

### Did PHPStan help you today? Did it make you happy in any way?

Every. single. day.

Thanks for this amazing tool!

ondrejmirtes:
That's a completely different code path, because it has to be implemented in phpdoc-parser.

PR #238, opened 2024-05-07
Add basic support for new-in-initializer

rvanvelzen:
Implementation for #173

### DIFF
--- a/src/Ast/ConstExpr/ConstExprNewNode.php
+++ b/src/Ast/ConstExpr/ConstExprNewNode.php
@@ -1,0 +1,34 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast\ConstExpr;
+
+use PHPStan\PhpDocParser\Ast\NodeAttributes;
+use function implode;
+
+class ConstExprNewNode implements ConstExprNode
+{
+
+	use NodeAttributes;
+
+	/** @var string */
+	public $class;
+
+	/** @var ConstExprNode[] */
+	public $arguments;
+
+	/**
+	 * @param ConstExprNode[] $arguments
+	 */
+	public function __construct(string $class, array $arguments)
+	{
+		$this->class = $class;
+		$this->arguments = $arguments;
+	}
+
+
+	public function __toString(): string
+	{
+		return 'new ' . $this->class . '(' . implode(', ', $this->arguments) . ')';
+	}
+
+}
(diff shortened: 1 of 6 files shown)


ondrejmirtes:
Do you see this somehow being useful for the analysis, or is it just because we don't have a different syntax for "let's make this parameter optional but not nullable"?

One thing this will impact are types resolved for generics and conditional types (that should be tested), but otherwise not much imho.

rvanvelzen:
Could you explain what this has to do with optional or non-nullable parameters?

> One thing this will impact are types resolved for generics and conditional types (that should be tested), but otherwise not much imho.

While this is true, for default values in regular code generics aren't taken into account either. So implementing this in PHPStan itself is just a few lines of code.

ondrejmirtes:
They are: https://phpstan.org/r/fe352005-91ed-4021-bd95-9d6df9f886a4 (and not generalized on bleeding edge https://phpstan.org/r/7fad672b-f8b0-4128-9282-d82a6599f2b6, unless in GenericObjectType).

ondrejmirtes:
> Could you explain what this has to do with optional or non-nullable parameters?

I don't really get the question, because the answer is "everything"? This syntax applies only to default value of parameters in `@method`. So I was interested how this enhances analysis in PHPStan. And the answer to that is that besides marking the parameter as optional, it will influence the result of generics and conditional types. Right?

rvanvelzen:
> They are: [phpstan.org/r/fe352005-91ed-4021-bd95-9d6df9f886a4](https://phpstan.org/r/fe352005-91ed-4021-bd95-9d6df9f886a4) (and not generalized on bleeding edge [phpstan.org/r/7fad672b-f8b0-4128-9282-d82a6599f2b6](https://phpstan.org/r/7fad672b-f8b0-4128-9282-d82a6599f2b6), unless in GenericObjectType).

Ah, I actually meant that the default value _itself_ is never generic: https://phpstan.org/r/dc1fee3f-b8de-448d-ba24-edc733770921

I think I understand what you're trying to figure out, and I think the short answer is: this is only to indicate that such a parameter has a default value, which is impossible otherwise right now. Not really relevant for analysis.

But this is nice and simple and mirrors the native syntax precisely. :)

ondrejmirtes:
I agree. I just wanted to show the test with generics should be added in PHPStan to show we're actually reading the default value correctly. And if the class in `new` is actually generic, we should also resolve it correctly. So `new Foo([1, 2, 3])` should result in correct `Foo<T>` same as in PHP code - we should probably construct PhpParser's `New_` node and just ask `Scope` about it...

ondrejmirtes (changes requested):
This needs some changes in Printer too. At least definitely a one in `$listInsertionMap` and maybe more.

Not all entries in that property are tested, but if you comment `ArrayShapeNode::class . '->items' => ', ',` you'll see the kind of test I'd like to see. It's somewhere around `$addItemsToArrayShape` in PrinterTest. Thanks!

PR #240, opened 2024-05-08
Add syntax for subtraction type

rvanvelzen:
This is a preliminary syntax implementation for subtraction type syntax.

Merging now isn't necessary because the work in phpstan-src is far from done. Any feedback is still appreciated though.

### DIFF
--- a/tests/PHPStan/Parser/TypeParserTest.php
+++ b/tests/PHPStan/Parser/TypeParserTest.php
@@ -26,6 +26,7 @@ use PHPStan\PhpDocParser\Ast\Type\NullableTypeNode;
 use PHPStan\PhpDocParser\Ast\Type\ObjectShapeItemNode;
 use PHPStan\PhpDocParser\Ast\Type\ObjectShapeNode;
 use PHPStan\PhpDocParser\Ast\Type\OffsetAccessTypeNode;
+use PHPStan\PhpDocParser\Ast\Type\SubtractionTypeNode;
 use PHPStan\PhpDocParser\Ast\Type\ThisTypeNode;
 use PHPStan\PhpDocParser\Ast\Type\TypeNode;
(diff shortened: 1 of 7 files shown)


ondrejmirtes:
1) We have to get rid of all Type inheritance before this is possible to do: https://phpstan.org/blog/why-is-instanceof-type-wrong-and-getting-deprecated
2) I'm not sure we need a new syntax for this and how it should look like. We could do this simply with `subtract-from<X, Y>` generics.

tests/PHPStan/Parser/TypeParserTest.php
```diff
@@ -276,6 +277,66 @@ public function provideParseData(): array
...
+				'  ~' . PHP_EOL .
+				'  int' . PHP_EOL .
+				')',
+				new SubtractionTypeNode(
+					new IdentifierTypeNode('string'),
+					new IdentifierTypeNode('int')
+				),
+			],
+			[
+				'string~int~float',
```
mvorisek, line 313: One non-parentheised expression should not allow more than one `~` to prevent associativity ambiguity as not standartized not needed.

As you can see, currently `float` is even missing in the parsed type.

tests/PHPStan/Parser/TypeParserTest.php
```diff
@@ -276,6 +277,66 @@ public function provideParseData(): array
...
+				new SubtractionTypeNode(
+					new IdentifierTypeNode('string'),
+					new IdentifierTypeNode('int')
+				),
+				Lexer::TOKEN_SUBTRACTION,
+			],
+			[
+				'(string&int)~float',
```
mvorisek, line 321: These cases should be tested:

- `A|B~C` (expected `A|(B~C)`)
- `A~B|C` (expected `(A~B)|C`)
- `A&B~C` (expected syntax error?)
- `A~B&C` (expected syntax error?)

PR #250, opened 2024-08-25
Support typing extra items in unsealed array shapes

### LINKED ISSUE #245: Add support for unsealed shapes with specified other elements' types
Psalm supports that: https://psalm.dev/r/cd2368bac0
PHPStan does not yet: https://phpstan.org/r/5a21f171-ec9a-467b-ad9e-6d1b1bd1de33

tscni:
Part of https://github.com/phpstan/phpstan/issues/8438
Resolves https://github.com/phpstan/phpdoc-parser/issues/245

### DIFF
--- a/src/Ast/Type/ArrayShapeNode.php
+++ b/src/Ast/Type/ArrayShapeNode.php
@@ -2,11 +2,8 @@
 
 namespace PHPStan\PhpDocParser\Ast\Type;
 
-use InvalidArgumentException;
 use PHPStan\PhpDocParser\Ast\NodeAttributes;
 use function implode;
-use function strlen;
-use function substr;
 
 class ArrayShapeNode implements TypeNode
 {
@@ -25,8 +22,11 @@ class ArrayShapeNode implements TypeNode
 	/** @var self::KIND_* */
 	public $kind;
 
-	/** @var GenericTypeNode|null */
-	public $extraItemType;
+	/** @var TypeNode|null */
+	public $extraKeyType;
+
+	/** @var TypeNode|null */
+	public $extraValueType;
 
 	/**
 	 * @param ArrayShapeItemNode[] $items
@@ -36,16 +36,15 @@ class ArrayShapeNode implements TypeNode
 		array $items,
 		bool $sealed = true,
 		string $kind = self::KIND_ARRAY,
-		?GenericTypeNode $extraItemType = null
+		?TypeNode $extraKeyType = null,
+		?TypeNode $extraValueType = null
 	)
 	{
 		$this->items = $items;
 		$this->sealed = $sealed;
 		$this->kind = $kind;
-		$this->extraItemType = $extraItemType;
-		if ($sealed && $extraItemType !== null) {
-			throw new InvalidArgumentException('An extra item type may only be set for an unsealed array shape');
-		}
+		$this->extraKeyType = $extraKeyType;
+		$this->extraValueType = $extraValueType;
 	}
 
 
@@ -55,8 +54,13 @@ class ArrayShapeNode implements TypeNode
 
 		if (! $this->sealed) {
 			$item = '...';
-			if ($this->extraItemType !== null) {
-				$item .= substr((string) $this->extraItemType, strlen((string) $this->extraItemType->type));
+			if ($this->extraValueType !== null) {
+				$extraTypes = [];
+				if ($this->extraKeyType !== null) {
+					$extraTypes[] = (string) $this->extraKeyType;
+				}
+				$extraTypes[] = (string) $this->extraValueType;
+				$item .= '<' . implode(', ', $extraTypes) . '>';
 			}
 			$items[] = $item;
 		}
--- a/src/Ast/Type/ArrayShapeNode.php
+++ b/src/Ast/Type/ArrayShapeNode.php
@@ -2,8 +2,11 @@
 
 namespace PHPStan\PhpDocParser\Ast\Type;
 
+use InvalidArgumentException;
 use PHPStan\PhpDocParser\Ast\NodeAttributes;
 use function implode;
+use function strlen;
+use function substr;
 
 class ArrayShapeNode implements TypeNode
 {
@@ -22,15 +25,27 @@ class ArrayShapeNode implements TypeNode
 	/** @var self::KIND_* */
 	public $kind;
 
+	/** @var GenericTypeNode|null */
+	public $extraItemType;
+
 	/**
 	 * @param ArrayShapeItemNode[] $items
 	 * @param self::KIND_* $kind
 	 */
-	public function __construct(array $items, bool $sealed = true, string $kind = self::KIND_ARRAY)
+	public function __construct(
+		array $items,
+		bool $sealed = true,
+		string $kind = self::KIND_ARRAY,
+		?GenericTypeNode $extraItemType = null
+	)
 	{
 		$this->items = $items;
 		$this->sealed = $sealed;
 		$this->kind = $kind;
+		$this->extraItemType = $extraItemType;
+		if ($sealed && $extraItemType !== null) {
+			throw new InvalidArgumentException('An extra item type may only be set for an unsealed array shape');
+		}
 	}
 
 
@@ -39,7 +54,11 @@ class ArrayShapeNode implements TypeNode
 		$items = $this->items;
 
 		if (! $this->sealed) {
-			$items[] = '...';
+			$item = '...';
+			if ($this->extraItemType !== null) {
+				$item .= substr((string) $this->extraItemType, strlen((string) $this->extraItemType->type));
+			}
+			$items[] = $item;
 		}
 
 		return $this->kind . '{' . implode(', ', $items) . '}';
--- a/src/Ast/Type/ArrayShapeNode.php
+++ b/src/Ast/Type/ArrayShapeNode.php
@@ -22,11 +22,8 @@ class ArrayShapeNode implements TypeNode
 	/** @var self::KIND_* */
 	public $kind;
 
-	/** @var TypeNode|null */
-	public $extraKeyType;
-
-	/** @var TypeNode|null */
-	public $extraValueType;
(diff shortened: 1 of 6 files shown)


src/Parser/TypeParser.php
```diff
@@ -858,6 +859,19 @@ private function parseArrayShape(TokenIterator $tokens, Ast\Type\TypeNode $type,
 
 			if ($tokens->tryConsumeTokenType(Lexer::TOKEN_VARIADIC)) {
 				$sealed = false;
+				$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
+				if ($tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_ANGLE_BRACKET)) {
+					$extraItemType = $this->parseGeneric(
+						$tokens,
+						$this->enrichWithAttributes(
+							$tokens,
+							new Ast\Type\IdentifierTypeNode($kind),
```
tscni: This and the necessary printing adjustments for it are very hacky. I don't like it at all and would rather add support for a `GenericNode` (matching the current grammar for `Generic` which doesn't have a type identifier)
But I'd prefer a second opinion before doing that.

So these are our options:
- Keep it like this and identify the type as `array` / `list`.
  The main benefit of this approach is that the type resolution in phpstan-src would require very little adjustments for this part, because it's exactly built for this. (https://github.com/phpstan/phpstan-src/blob/04b840344a5f403e9a48120d7b4dccb278cab2d2/src/PhpDoc/TypeNodeResolver.php#L640-L669)
  But I don't mind doing more work there at all.
  
  The disadvantage: Semantically & grammatically it's not accurate and it requires workarounds for printing & string conversions.
- Use an empty string as the type identity
  The main benefit is that it works quite well in this repository and no printing  & string conversions workarounds are necessary.
  
  The disadvantage: Semantically & grammatically it still feels inaccurate
- Add a new `GenericNode` (or similar), basically just `GenericTypeNode` without the base type.
  This would be the most accurate because there really is no base type.

  The disdvantage: Ideally this would then be integrated into `GenericTypeNode`, but for backward compatibility reasons that won't be as simple / currently not possible, slightly raising the complexity when the two interact.

tests/PHPStan/Parser/TypeParserTest.php
tscni, line 1: I intentionally ignored different generic variances in the tests. Due to the copy-on-write behavior from arrays, these should never be relevant, but I haven't fully thought this through yet and might be wrong.

Adding some error cases wouldn't hurt either.

src/Ast/Type/ArrayShapeNode.php
```diff
@@ -22,15 +25,27 @@ class ArrayShapeNode implements TypeNode
...
 	/**
 	 * @param ArrayShapeItemNode[] $items
 	 * @param self::KIND_* $kind
 	 */
-	public function __construct(array $items, bool $sealed = true, string $kind = self::KIND_ARRAY)
+	public function __construct(
+		array $items,
+		bool $sealed = true,
+		string $kind = self::KIND_ARRAY,
+		?GenericTypeNode $extraItemType = null
```
tscni: Different naming suggestions are very welcome, though it's probably fine

src/Parser/TypeParser.php
```diff
@@ -858,6 +859,19 @@ private function parseArrayShape(TokenIterator $tokens, Ast\Type\TypeNode $type,
 
 			if ($tokens->tryConsumeTokenType(Lexer::TOKEN_VARIADIC)) {
 				$sealed = false;
+				$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
+				if ($tokens->isCurrentTokenType(Lexer::TOKEN_OPEN_ANGLE_BRACKET)) {
+					$extraItemType = $this->parseGeneric(
```
tscni: I thought about adjusting the grammar, but there currently doesn't seem to be a big focus on keeping the grammar maintained. There's e.g. no representation for `TOKEN_PHPDOC_EOL` at all, `ArrayShape` is already out of date etc. 
I could work on all of that, but again, if there's not much interest in keeping it maintained that doesn't quite seem worthwhile.

Ultimately this leads to the question what the point of the grammar is. 
Is it mainly for documentation / easier understanding of the parser logic? Or is there a goal to generate more documentation based on it?
Is the end goal to move to a generated parser?
Should it serve for third party consumers?

I suppose ideally the tests would somehow verify that the parse results match the current grammar.

tscni (commented):
This is the first iteration with some open questions. I would mark it as a draft, but somehow I feel that it'd get less feedback that way.

src/Ast/Type/ArrayShapeNode.php
```diff
@@ -22,15 +25,27 @@ class ArrayShapeNode implements TypeNode
 	/** @var self::KIND_* */
 	public $kind;
 
+	/** @var GenericTypeNode|null */
+	public $extraItemType;
```
ondrejmirtes: Hey, I'd rather have a different design:

```php
/** @var TypeNode|null */
public $extraKeyType;

/** @var TypeNode|null */
public $extraValueType;

// ...

/**
 * @phpstan-assert-if-false !null $this->extraKeyType
 * @phpstan-assert-if-false !null $this->extraValueType
 */
public function isSealed(): bool
{
}

/**
 * @phpstan-assert-if-true !null $this->extraKeyType
 * @phpstan-assert-if-true !null $this->extraValueType
 */
public function isUnsealed(): bool
{
}
```

The GenericTypeNode here is too broad, it allows for invalid AST with types count that isn't 2.

ondrejmirtes (changes requested):

src/Ast/Type/ArrayShapeNode.php
tscni: I mainly went for `GenericTypeNode` to keep consistency with regular array definitions (e.g. `array<string, string, string>` works fine on a doc parser level), but I'll happily make it more restrictive.

The assertions on the other hand seem slightly concerning, specifically for unsealed shapes without extra key/value types.
- Given `array{...}` 
- We can internally represent that as `array{...<array-key, mixed>}`
- And for printing / string conversions output that as `array{...}` again (also to keep backward compatibility)
- But this also means that an explicit `array{...<array-key, mixed>}` would also result in an output of `array{...}`, unless we add a distinction for "not set"
- Also, `list{...}` would then internally be represented as `list{...<mixed>}` (internally with an irrelvant `int` key type)

I don't mind doing that, but is the complexity really more suitable in the parser vs. handling it entirely on the type resolution side? (I suppose there's a fine line between the two)

ondrejmirtes: I don't want to allow `array{...}` without any items, we should throw a parse error for that. People should use a general array instead.

tscni: It's currently allowed. Is that breaking change ok? (see the PR that introduced it https://github.com/phpstan/phpdoc-parser/pull/169)

Note that my concerns also apply for e.g. `array{foo: string, ...}` and `list{string, ...}`

tscni: Of note: Psalm allows `array{...}` & `list{...}` and, like `array{}` & `list{}`, resolves them to `array<never, never>`: https://psalm.dev/r/7840e76832
Basically meaning that `array{...}` is still considered sealed by Psalm. (feels like a bug)

In PHPStan they're similarly all resolved to `array{}` at the moment: https://phpstan.org/r/69df8759-e426-4d5d-a8d3-83d26f973624

ondrejmirtes: Alright, I get it. As for the AST in phpdoc-parser, I guess we need to throw away my idea for the methods with `@phpstan-assert`. We'll keep the `public $sealed` which will still mean "are there `...` in the array shape?". And then we should have a new AST object which will contain both key type nad value type.

So there will be some kind of "ArrayShapeUnsealedTypes|null" property in ArrayShapeNode with `$keyType` and `$valueType`. Which will make it impossible to represent impossible states in the object structure.

The only impossible state to represent would be `sealed=true` and `unsealedType !== null` but I'm fine to live with that.

As for the PHPStan typesystem representation, we can do the same thing we're doing to `array<X>` vs. `array<array-key, X>`.

`array{foo: X, ...}` would be represented with benevolent union type in the place of unsealed key type, and implicit mixed type in place of the value. So we'd still know what to do in `toPhpDocNode()`.

`array<foo: X, ...<array-key, mixed>` would still have the benevolent union type, but the mixed would be explicit, so we'd know what to do in `toPhpDocNode()`.

Thanks!

tscni: > The only impossible state to represent would be sealed=true and unsealedType !== null but I'm fine to live with that.

There's also the impossible state of `$kind === 'list' && $unsealedType->keyType !== null`, but that's it.

I've made the necessary adjustments.

src/Ast/Type/ArrayShapeUnsealedTypeNode.php
```diff
@@ -0,0 +1,33 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\PhpDocParser\Ast\Type;
+
+use PHPStan\PhpDocParser\Ast\NodeAttributes;
+use function sprintf;
+
+class ArrayShapeUnsealedTypeNode implements TypeNode
```
ondrejmirtes: This should be just a Node, not TypeNode. It doesn't make sense standalone.

ondrejmirtes (changes requested):
FYI I pushed a commit with a few tests I was missing :)

src/Ast/Type/ArrayShapeUnsealedTypeNode.php
tscni: Makes sense. `ArrayShapeItemNode` shouldn't have been one either then I guess, oh well.
Done.

tscni:
> FYI I pushed a commit with a few tests I was missing :)

Thanks :)
I added some more cases for list shapes

ondrejmirtes:
Awesome, thank you!

PR #252, opened 2024-09-06
Add test cases for quote aware array/object keys

### LINKED ISSUE #251: Key should be quoted when it contains spaces in ArrayShapeItemNode
I noticed something odd.

```php
$input = "array{states: array{'Hello World': int}}";
$expected = "array{states: array{'Hello World': int}}";
$lexer = new Lexer();
$constExprParser = new ConstExprParser(true, true);
$typeParser = new TypeParser($constExprParser);
$phpDocParser = new PhpDocParser($typeParser, $constExprParser);

$tokens = new TokenIterator($lexer->tokenize($input));
$phpDocNode = $phpDocParser->parseTagValue($tokens, '@var');

$this->assertInstanceOf(VarTagValueNode::class, $phpDocNode);
$this->assertInstanceOf(ArrayShapeNode::class, $phpDocNode->type);

$printer = new Printer();
$this->assertSame($expected, $printer->print($phpDocNode));
```

> Failed asserting that two strings are identical.
> Expected :'array{states: array{'Hello World': int}}'
> Actual   :'array{states: array{Hello World: int}}'

As you can see, the key `Hello World` is not quoted, while it should.

When I try to feed the wrong PHPDoc into `parseTagValue` it returns `InvalidTagValueNode`.

Am I doing something wrong, or is this a bug that I can try to fix 😊 ?

ruudk:
See #251

### DIFF
--- a/tests/PHPStan/Printer/PrinterTest.php
+++ b/tests/PHPStan/Printer/PrinterTest.php
@@ -2010,6 +2010,36 @@ class PrinterTest extends TestCase
 			),
 			'self::TYPES[int]',
 		];
+		yield [
+			new ArrayShapeNode([
+				new ArrayShapeItemNode(
+					new IdentifierTypeNode('name'),
+					false,
+					new IdentifierTypeNode('string')
+				),
+				new ArrayShapeItemNode(
+					new QuoteAwareConstExprStringNode('Full Name', QuoteAwareConstExprStringNode::SINGLE_QUOTED),
+					false,
+					new IdentifierTypeNode('string')
+				),
+			]),
+			"array{name: string, 'Full Name': string}",
+		];
+		yield [
+			new ObjectShapeNode([
+				new ObjectShapeItemNode(
+					new IdentifierTypeNode('name'),
+					false,
+					new IdentifierTypeNode('string')
+				),
+				new ObjectShapeItemNode(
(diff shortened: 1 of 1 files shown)


ondrejmirtes (changes requested):
Hi, the right test to test this is to provide the PHPDoc, parse it, assert the AST, then print it, and then try to parse it again, and then assert the AST again.

Something like PrinterTest::testPrintType already does.

Here I'm not sure how the keyName is represented. There are definitely more situations where it can happen besides a space.

Ideally the keyName should be represented as QuoteAwareConstExprStringNode which already carries the type of quotes and is able to print itself correctly.

ruudk:
I agree. Will try it out

ruudk:
I added the test cases and they work perfectly fine.

~That's interesting.~

~I think the issue is in PHPStan:
https://github.com/phpstan/phpstan-src/blob/8fd91a40fa9064378b18a31c081663d6f54494e6/src/Type/Constant/ConstantArrayType.php#L1648-L1689~

~This never creates `QuoteAwareConstExprStringNode`.~

ruudk:
I'm sorry for this, but the problem was that my `TypeParser` did not have `$quoteAwareConstExprString` enabled. While my `ConstExprParser` did. I didn't notice I had to configure it in 2 locations.

ruudk:
Is there value in these 2 test cases? Or do we close the PR?

ondrejmirtes:
Thank you!

PR #254, opened 2024-09-20
PhpDocParser: support template type lower bounds

jiripudil:
I'd like to take a shot at template type lower bounds (phpstan/phpstan#5179). This PR adds support for parsing lower bounds.

As I've reasoned in https://github.com/phpstan/phpstan/issues/5179#issuecomment-1529134067: while I don't think it's particularly useful, the implementation that I envision in phpstan/phpstan-src will be able to account for a template having both bounds, regardless of whether it's allowed in the parser or not. Thus, I don't see any technical reason _not_ to allow it.


ondrejmirtes:
Awesome, thank you!

PR #255, opened 2024-10-09
Fix IntegrationPrinterWithPhpParserTest

staabm:
since PHPParser 5.3 this integration test fails.

it overlapps with a upstream implemenation of indentation-printing which was added with https://github.com/nikic/PHP-Parser/commit/e50c67b7a94d2d028f8792f462e31ba4ed9f2680

the PR removes the phpdoc-parser custom build indentation printer and re-uses the upstream indentation printer instead.


staabm:
on my windows box for some reason I cannot fix CS here (creates a lot of unrelated changes).
@ondrejmirtes would be great you could fix cs/phpstan-baseline, in case the fix itself works for you

tests/PHPStan/Printer/PhpPrinterIndentationDetectorVisitor.php
```diff
@@ -71,7 +72,7 @@ public function enterNode(Node $node)
 			$this->indentCharacter = $char;
 			$this->indentSize = $size;
 
-			return NodeTraverser::STOP_TRAVERSAL;
+			return NodeVisitor::STOP_TRAVERSAL;
```
staabm, line 74: drive-by deprecation fix

staabm:
in case we agree on this one, I can have a look at simple-downgrader 2.x

ondrejmirtes:
Awesome, thank you!

> I can have a look at simple-downgrader 2.x

Yes please :)

PR #253, opened 2024-09-18
Support for `@pure-unless-callable-is-impure`

staabm:
required for https://github.com/phpstan/phpstan/issues/11100

PR #259, opened 2024-11-18
Support for `@pure-unless-parameter-passed`

staabm:
refs https://github.com/phpstan/phpstan/issues/11884#issuecomment-2426297188

### DIFF
--- a/src/Ast/PhpDoc/PhpDocNode.php
+++ b/src/Ast/PhpDoc/PhpDocNode.php
@@ -127,6 +127,17 @@ class PhpDocNode implements Node
 	}
 
 	/**
+	 * @return PureUnlessParameterIsPassedTagValueNode[]
+	 */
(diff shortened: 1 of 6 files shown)


ondrejmirtes:
We already have other similar implementation lined up. There's `@pure-unless-callable-is-impure` which is implemented in the parser (https://github.com/phpstan/phpdoc-parser/pull/253) but not yet in PHPStan (https://github.com/phpstan/phpstan-src/pull/3482).

It's not a great look when there's a tag understood by the parser but not by the analyser. I'd like to minimize that coming into the future. It'd be nice to also have the implementation ready in phpstan-src before merging the parser countepart.

I will look at this again when https://github.com/phpstan/phpstan-src/pull/3482 is merged and the implementation for `@pure-unless-parameter-passed` is ready in phpstan-src.

zonuexe:
@ondrejmirtes https://github.com/phpstan/phpstan-src/pull/6018 is ready; could you please consider merging this PR?

ondrejmirtes:
Merged and tagged.

Issue #261, opened 2025-01-15
Inline tags cause processing of subsequent tags to be discarded


andrewnicols:
We have some (mostly legacy) code which makes use of inline PHPDoc tags within other tags, for example the `@copyright` line in the following example contains an inline `@link` tag:

```php
/**
 * Unit tests for stored_progress_bar_cleanup
 *
 * @package   core
 * @copyright 2024 onwards Catalyst IT EU {@link https://catalyst-eu.net}
 * @author    Mark Johnson <mark.johnson@catalyst-eu.net>
 * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
 * @covers    \core\task\stored_progress_bar_cleanup_task
 */
```

When Rector's BetterDocsParser processes this using `phpstan\phpdoc-parser` I'm seeing that the processing stops at the `@copyright` line just before the inline tag.

From a quick dive into the code it looks like this may be an issue with `\PHPStan\PhpDocParser\Parser\PhpDocParser::parseOptionalDescriptionAfterDoctrineTag`, which correctly identifies the `@link` tag, but because it is a `Ast\PhpDoc\GenericTagValueNode`, it stops processing the rest of the docblock tags any further.

https://github.com/phpstan/phpdoc-parser/blob/04c8de0d350731bccfa216a5b5e9b3e1c26e4e24/src/Parser/PhpDocParser.php#L255-L262

As a result the only tags returned are:

```php
 @package  => "core"
 @copyright => "2024 onwards Catalyst IT EU "
```

Given there is no authoritative PHPDoc reference it's hard to determine what the correct behaviour is here.

The phpDocumentor manual states that the `inline {@link}` tag is to:
> Display a link to a URL, or link to an element's documentation in the the flow of descriptive text
https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.inlinelink.pkg.html

The FIG proposal does not provide any guidance that I can see on the use of inline tags, beyond that fact that they exist and that `@link` is a valid inline tag.
https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md

However the format of the copyright tag is a bit uncertain.

The phpDocumentor manual states that it is used unaltered:
> phpDocumentor will display the copyright string unaltered.
https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.copyright.pkg.html

[...]

PR #262, opened 2025-01-16
Test inline `@link` in Lexer

andrewnicols:
This is a followup to #261. Whilst the parsing of the tags themselves is correct in phpstan, Rector uses the PHPStan Lexer to tokenize the values. It seems that the tokenisation bundles the closing curly brace into the previous text content unless there is whitespace between them.

This test demonstrates that issue.

This issue is also present for opening/closing angle brackets, and parens. It seems that the OTHER token is too greedy.

It may be that this is the intended behaviour of the Lexer but I haven't been able to find any documentation or information to confirm either way.

https://github.com/rectorphp/rector/issues/8977 has additional info on the issue and how it is encountered.

### DIFF
--- a/tests/PHPStan/Parser/PhpDocParserTest.php
+++ b/tests/PHPStan/Parser/PhpDocParserTest.php
@@ -6989,6 +6989,23 @@ Finder::findFiles('*.php')
 	}
 
 	/**
+	 * @dataProvider provideLexerData
+	 */
+	public function testClosingBraceInlineTag(
+		string $input,
+		array $expectedTokens,
+	): void {
+		$tokens = new TokenIterator($this->lexer->tokenize($input));
+		$this->assertEquals($expectedTokens, array_map(
+			fn ($token) => [
+				$token[Lexer::TYPE_OFFSET],
+				$token[Lexer::VALUE_OFFSET],
+			],
+			$tokens->getTokens(),
+		));
+	}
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
A little personal rant here: I appreciate what the Rector guys do, but they always implement hacky workarounds for bugs or missing features in PHPStan and phpdoc-parser. They are the heaviest users of PHPStan's internals but they rarely open any issues or pull requests.

If they properly reported their needs, I could address them here and they could use the parser in a clean way, without the need to descend into the lexer.

As an example, slevomat/coding-standard does this right. When I implemented the format-preserving printer, slevomat/coding-standard adopted in saved literally thousands of lines of code in their repository: https://github.com/slevomat/coding-standard/pull/1577

If Rector collaborated more and used PHPStan in a proper way without workarounds, it'd be easier for everyone.

andrewnicols:
Sorry @ondrejmirtes, I can only agree. I'm just a frustrated user trying to work out why my rules aren't being applied properly and trying to take the time to not only report a bug but try and find a fix at the same time. I've only started using Rector in the past few days.

Do you think that this is an expected behaviour in the Lexer then?

ondrejmirtes:
I'd say the specific representation in the lexer doesn't matter as long as the parser consumes it and produces proper AST.

Issue #258, opened 2024-11-12
Type parsing failes if an empty line is in between


M393:
Had an empty line between properties of a large object, but phpdoc-parser cannot parse it.
https://phpstan.org/r/c23f2e88-4908-481c-9a14-f1caa68b8e5e

DaDeather:
I've opened a PR to support this https://github.com/phpstan/phpdoc-parser/pull/263

ondrejmirtes:
Implemented https://github.com/phpstan/phpdoc-parser/pull/263

Issue #183, opened 2023-04-05
Union operator on a new line causes parsing error


internalsystemerror:
Original issue: https://github.com/phpstan/phpstan/issues/9019

When we have a type defined such as

```php
/**
 * @phpstan-type FactoriesConfigurationType = array<
 *      string,
 *      (class-string<Factory\FactoryInterface>|Factory\FactoryInterface)
 *      |callable(ContainerInterface,?string,?array<mixed>|null):object
 * >
 */
```

We get an error `Unexpected token "|", expected '>'`. According to the original issue the problem is caused by this package.

DaDeather:
I've opened a PR to support this https://github.com/phpstan/phpdoc-parser/pull/263

And as a side note: there was also an error in your callable definition in regards to `?array<mixed>|null`.

ondrejmirtes:
Implemented https://github.com/phpstan/phpdoc-parser/pull/263

PR #263, opened 2025-02-11
Allow new lines in between declarations and also allow multiline union and intersection definition for array shapes

DaDeather:
Takes care of:
- https://github.com/phpstan/phpdoc-parser/issues/258
- https://github.com/phpstan/phpdoc-parser/issues/183
- and also allows multiline declarations with union and intersections

### DIFF
--- a/src/Parser/TypeParser.php
+++ b/src/Parser/TypeParser.php
@@ -40,15 +40,42 @@ class TypeParser
 		} else {
 			$type = $this->parseAtomic($tokens);
 
-			if ($tokens->isCurrentTokenType(Lexer::TOKEN_UNION)) {
-				$type = $this->parseUnion($tokens, $type);
+			$tokens->pushSavePoint();
+			$tokens->skipNewLineTokens();
 
-			} elseif ($tokens->isCurrentTokenType(Lexer::TOKEN_INTERSECTION)) {
-				$type = $this->parseIntersection($tokens, $type);
+			try {
+				$enrichedType = $this->enrichTypeOnUnionOrIntersection($tokens, $type);
+
+			} catch (ParserException $parserException) {
+				$enrichedType = null;
+			}
+
+			if ($enrichedType !== null) {
+				$type = $enrichedType;
+				$tokens->dropSavePoint();
+
+			} else {
+				$tokens->rollback();
+				$type = $this->enrichTypeOnUnionOrIntersection($tokens, $type) ?? $type;
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


src/Parser/TypeParser.php
```diff
@@ -409,8 +409,13 @@ public function parseGeneric(TokenIterator $tokens, Ast\Type\IdentifierTypeNode
 		$variances = [];
 
 		$isFirst = true;
-		while ($isFirst || $tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA)) {
-			$tokens->tryConsumeTokenType(Lexer::TOKEN_PHPDOC_EOL);
+		while (
+			$isFirst
+			|| $tokens->tryConsumeTokenType(Lexer::TOKEN_COMMA)
+			|| $tokens->tryConsumeTokenType(Lexer::TOKEN_UNION)
+			|| $tokens->tryConsumeTokenType(Lexer::TOKEN_INTERSECTION)
```
ondrejmirtes: This doesn't make sense to me. The code makes it look like `|` and `&` could be used in place of `,` in a generic type.

ondrejmirtes (changes requested):

src/Parser/TypeParser.php
ondrejmirtes: Feel free to remove this part from the PR and submit it separately, the rest is solid.

DaDeather: You're totally right it was totally the wrong place.

I moved it accordingly and also adjusted the tests which if I would have had a better look would have already made clear that the parsed information would therefore be wrong :see_no_evil:.

I had to adjust the `TypeParser::parse` by using savepoints to check whether the union &/ intersection is valid to when being parsed to rollback when somethings feels off due to the information outputted afterwards not being shown "right".

So e.g. parsing `/** @return A & B | C */` would had shown a error with partial information missing the beginning showing off a `& B | C` instead. (applied on test case `invalid with type and disallowed description start token (2)`)

So if you by any chance have a better idea or solution to making this happen I'd be happy about any improvement here.

Otherwise this should now fulfill it's purpose allowing multiline type declarations and other multiline operations that may have led to an error before as stated with the tests and the linked issues that this would solve.

Let me know or feel free to make adjustments yourself if there's something we should do here :+1:.

Otherwise thanks for the effort you're putting into all of this :heart:!

ondrejmirtes:
Nice, thank you!

PR #264, opened 2025-02-14
TypeParser - support comments at EOL with `//`

### LINKED ISSUE #184: Multiline array shapes with 'comments'

ondrejmirtes:
Closes https://github.com/phpstan/phpdoc-parser/issues/184

### DIFF
--- a/src/Lexer/Lexer.php
+++ b/src/Lexer/Lexer.php
@@ -51,6 +51,8 @@ class Lexer
 	public const TOKEN_NEGATED = 35;
 	public const TOKEN_ARROW = 36;
 
+	public const TOKEN_COMMENT = 37;
+
 	public const TOKEN_LABELS = [
 		self::TOKEN_REFERENCE => '\'&\'',
 		self::TOKEN_UNION => '\'|\'',
@@ -66,6 +68,7 @@ class Lexer
 		self::TOKEN_OPEN_CURLY_BRACKET => '\'{\'',
 		self::TOKEN_CLOSE_CURLY_BRACKET => '\'}\'',
 		self::TOKEN_COMMA => '\',\'',
+		self::TOKEN_COMMENT => '\'//\'',
 		self::TOKEN_COLON => '\':\'',
 		self::TOKEN_VARIADIC => '\'...\'',
 		self::TOKEN_DOUBLE_COLON => '\'::\'',
@@ -160,6 +163,7 @@ class Lexer
 			self::TOKEN_CLOSE_CURLY_BRACKET => '\\}',
 
 			self::TOKEN_COMMA => ',',
+			self::TOKEN_COMMENT => '\/\/[^\\r\\n]*(?=\n|\r|\*/)',
 			self::TOKEN_VARIADIC => '\\.\\.\\.',
 			self::TOKEN_DOUBLE_COLON => '::',
 			self::TOKEN_DOUBLE_ARROW => '=>',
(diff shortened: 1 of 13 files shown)


src/Parser/TokenIterator.php
```diff
@@ -19,11 +20,16 @@ class TokenIterator
...
+	private array $comments = [];
+
 	/** @var int[] */
 	private array $savePoints = [];
 
 	/** @var list<int> */
-	private array $skippedTokenTypes = [Lexer::TOKEN_HORIZONTAL_WS];
+	private array $skippedTokenTypes = [
+		Lexer::TOKEN_HORIZONTAL_WS,
+		Lexer::TOKEN_COMMENT];
```
tscni, line 32: Don't forget about `addEndOfLineToSkippedTokens()` and `removeEndOfLineFromSkippedTokens()`.

src/Lexer/Lexer.php
```diff
@@ -160,6 +163,7 @@ private function generateRegexp(): string
 			self::TOKEN_CLOSE_CURLY_BRACKET => '\\}',
 
 			self::TOKEN_COMMA => ',',
+			self::TOKEN_COMMENT => '\/\/[^\\r\\n]*(?=\n|\r|\*/)',
```
mvorisek, line 166:
```suggestion
			self::TOKEN_COMMENT => '\/\/(?:(?!\*/).)*',
```

ondrejmirtes: Does not work.

mvorisek:
Did you considered using `--` instead of `//`?

This way `-* xx *-` (like `/* xx */` in code) would be complementary possible for inline comments.

PR #265, opened 2025-02-16
Allow multiple new lines in union and intersection declaration

DaDeather:
I just noticed that my latest implementation only allowed a "single" new line in between the type declaration for union and intersection types.

See:
https://phpstan.org/r/107386a9-94b0-4dc1-bc22-ac2d6db189c4

This change would allow multiple new lines between those to be handled correctly.

### DIFF
--- a/tests/PHPStan/Parser/PhpDocParserTest.php
+++ b/tests/PHPStan/Parser/PhpDocParserTest.php
@@ -60,6 +60,7 @@ use PHPStan\PhpDocParser\Ast\Type\ConditionalTypeNode;
 use PHPStan\PhpDocParser\Ast\Type\ConstTypeNode;
 use PHPStan\PhpDocParser\Ast\Type\GenericTypeNode;
 use PHPStan\PhpDocParser\Ast\Type\IdentifierTypeNode;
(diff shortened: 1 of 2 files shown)


tests/PHPStan/Parser/PhpDocParserTest.php
```diff
@@ -4211,6 +4212,176 @@ public function provideMultiLinePhpDocData(): iterable
 			]),
 		];
 
+		yield [
+			'Multiline PHPDoc with multiple new line within union type declaration',
+			'/**' . PHP_EOL .
```
mvorisek, line 4218: nowdoc can be used here for more readability

ondrejmirtes: Like here https://github.com/phpstan/phpdoc-parser/commit/dc7c4699f97407ab8cc99413cf32670d2f33a992

ondrejmirtes: But maybe not because that's only in PrinterTest.

ondrejmirtes:
Thank you!

DaDeather:
> Thank you!

Sure, happy to help! But rather thank you for also fixing the tests that failed in this regard!

PR #273, opened 2025-07-12
Add support for `psalm-inheritors` and `phpstan-sealed` tag

VincentLanglet:
Needed to implements the following feature on phpstan side
https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-inheritors


ondrejmirtes:
I’d prefer a different name for the PHPStan tag.

some keywords to consider:

* subclass
* subtype
* implementation
* allowed subclass
* Allowed subtype

What are your suggestions?

VincentLanglet:
I checked all the possibility and psalm supports them
https://psalm.dev/r/8a398e4a93

> I’d prefer a different name for the PHPStan tag.
> 
> some keywords to consider:
> 
> * subclass
> * subtype
> * implementation
> * allowed subclass
> * Allowed subtype
> 
> What are your suggestions?

Since it works for `interface` too ; I would not use the `subclass` word.
Since it works for `extends` too ;  I would not use the `implementation` word.

So I would say something like
```
@phpstan-subtypes
@phpstan-allowed-subtypes
@phpstan-restricted-subtypes
@phpstan-children
@phpstan-allowed-children
@phpstan-restricted-children
```
`@phpstan-subtypes` might be the simplest/shortest.

But if we look at other languages (and the related php rfc https://wiki.php.net/rfc/sealed_classes), seems adding a 
`@psalm-inheritors` means
- The class/interface is sealed
- The permitted extensions are the one listed

Therefor maybe something like `@phpstan-sealed Foo|Bar` or `@phpstan-permits Foo|Bar` ?

Or `@phpstan-permitted-subtypes` then ?

Does anything satisfy you @ondrejmirtes ?

ondrejmirtes:
`@phpstan-allowed-subtypes` or `@phpstan-sealed` might be the best in line with https://phpstan.org/developing-extensions/allowed-subtypes and https://github.com/jiripudil/phpstan-sealed-classes. I like the sealed better.

ondrejmirtes:
Of course Phpstan should understand psalm inheritors too.

VincentLanglet:
> I like the sealed better.

Done with `@phpstan-sealed` then

VincentLanglet:
The 
```
[Test Slevomat Coding Standard / Tests (8.4) (pull_request)](https://github.com/phpstan/phpdoc-parser/actions/runs/16237681772/job/45849904924?pr=273)Failing after 26s
```
is failing because of php deprecation on PHPUnit 10.5.11.

The dev-dependency was bumped on slevomat coding standard in commit 4747e312a24d4827b1d090b5908c7d5b16cf0580
Should I bump the commit in the github actions ?

ondrejmirtes:
@VincentLanglet Yes, in a separate PR please.

ondrejmirtes:
Thank you! Please send a PR to phpstan-src as well.

PR #275, opened 2025-07-28
Support constant fetch in array key

VincentLanglet:
I'm trying to solve 
https://github.com/phpstan/phpstan/issues/5773
and
https://github.com/phpstan/phpstan/issues/6989

By allowing the syntax `array{Foo::BAR: int}`.

Not fully sure if I'm doing it the right way.


ondrejmirtes:
Thank you! Please add support in phpstan-src.

PR #277, opened 2025-10-07
Introduce all-methods-pure and all-methods-impure tag

VincentLanglet:


ondrejmirtes:
I'm sorry, but I don't think these PHPDoc tags need any change in phpdoc-parser. `@phpstan-pure` and `@phpstan-impure` also don't have any references here because (I believe) they are represented with GenericTagValueNode (they just support a description, nothing else that needs to be parsed).

VincentLanglet:
Indeed cf https://github.com/phpstan/phpstan-src/pull/4422

PR #293, opened 2026-02-20
feat: add support for property access in conditional types

brambaud:
See https://github.com/phpstan/phpstan/issues/7545

From what I understood, we first need phpdoc-parser to be able to parse things like `@return ($this->t is T::A ? string : null)`  and then we'll need follow-up PR to phpstan/phpstan-src.

It also supports `parent::$value`, `self::$value` and `static::$value` because I thought it would be better to be consistent. If this is too much I understand.


brambaud:
Should we directly support null-safe operator in doc too?

I chose to let it for another time but maybe it does not makes sense without it

PR #298, opened 2026-05-07
Add `TOKEN_PHPDOC_INLINE_TAG`

kubawerlos:
Ref. https://github.com/phpstan/phpstan-src/pull/5583#discussion_r3199068032

// cc @staabm


ondrejmirtes (changes requested):
If someone modifies inlineTags in PhpDocTextNode, we should be able to properly support it in the format-preserving printer. Please add some cases to dataPrintFormatPreserving.

PR #310, opened 2026-08-30
Run the tests written for `@pure-unless-parameter-passed`

ondrejmirtes:
`providePureUnlessParameterIsPassedTagsData` was written for `@pure-unless-parameter-passed` in fb19eed, but was never named among the `@dataProvider` lines `testParse` is run with. Its two cases had never run.

They do not pass. `PureUnlessParameterIsPassedTagValueNode` is the one concrete node of `src/Ast/PhpDoc` without a `__set_state()`, and `testParse` writes every node out with `var_export()` and reads it back:

```
Error: Call to undefined method PHPStan\PhpDocParser\Ast\PhpDoc\PureUnlessParameterIsPassedTagValueNode::__set_state()
```

So this names the provider and gives the node the `__set_state()` its siblings have — `PureUnlessCallableIsImpureTagValueNode`'s, word for word.

Found while writing the grammar-sync test of #309, which reads the inputs of every other test out of the providers they are written with, and so walks over providers nothing else runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GavWEzia8ZVUYEe6JEde9i

### DIFF
--- a/src/Ast/PhpDoc/PureUnlessParameterIsPassedTagValueNode.php
+++ b/src/Ast/PhpDoc/PureUnlessParameterIsPassedTagValueNode.php
@@ -26,4 +26,18 @@ class PureUnlessParameterIsPassedTagValueNode implements PhpDocTagValueNode
 		return trim("{$this->parameterName} {$this->description}");
 	}
 
+	/**
+	 * @param array<string, mixed> $properties
+	 */
+	public static function __set_state(array $properties): self
+	{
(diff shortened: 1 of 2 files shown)


staabm:
Nice catch. Thank you. 

I guess PHPStan would have caught it as a unused method in case shipmonk/dead-code-detector would be included in this package?

ondrejmirtes:
Yes, it should catch that

